PR #10, opened 2021-10-19
Feat: Configure plugin defaults

OndraM:
Add some plugin defaults, as discussed with product manager.

### DIFF
--- a/src/LmcCookieConsentManager.js
+++ b/src/LmcCookieConsentManager.js
@@ -12,9 +12,15 @@ const LmcCookieConsentManager = (options = defaultOptions) => {
 
   cookieconsent.run({
     current_lang: currentLang,
+    auto_language: true,
     theme_css: themeCss,
     cookie_name: 'lmc_ccm',
     cookie_expiration: 365,
+    use_rfc_cookie: true,
+    autorun: true,
+    delay: 0,
+    force_consent: false,
+    hide_from_bots: true, // To be hidden also from Selenium
     gui_options: {
       consent_modal: {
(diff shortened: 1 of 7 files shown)


src/LmcCookieConsentManager.js
```diff
@@ -12,9 +12,15 @@ const LmcCookieConsentManager = (options = defaultOptions) => {
 
   cookieconsent.run({
     current_lang: currentLang,
+    auto_language: true,
     theme_css: themeCss,
     cookie_name: 'lmc_ccm',
     cookie_expiration: 365,
+    use_rfc_cookie: true,
+    autorun: true,
```
literat, line 20: This is `true` by default. I suggest to remove it.

src/LmcCookieConsentManager.js
```diff
@@ -12,9 +12,15 @@ const LmcCookieConsentManager = (options = defaultOptions) => {
...
+    delay: 0,
```
literat, line 21: This is `0` by default. I suggest to remove it.

src/LmcCookieConsentManager.js
```diff
@@ -12,9 +12,15 @@ const LmcCookieConsentManager = (options = defaultOptions) => {
...
+    force_consent: false,
```
literat, line 22: This is `false` by default. I suggest to remove it.

src/LmcCookieConsentManager.js
literat: Ok.

PR #11, opened 2021-10-20
Chore: Run all check with single npm script

literat:


OndraM:
The usecase I meant is for a developer - to be able to run some one-command script, which will run all the checks, lints and tests locally, on-demand - before they fail on the CI server, providing me faster feedback.

So it should not be called `ci`, imho, this would confuse me (i would expect this to be a script which should be run on CI server). I would suggest something like `yarn check` or - for example as we are used from PHP packages - `yarn all` (meaning "run all checks"). What do you think?

OndraM (approved):
I'm OK with this - `start` and `test` is understandable for me. All the other are up to you / @adamkudrna . 

However would be great if we can get this merged during monday - so cc @adamkudrna could you have a look?

package.json
```diff
@@ -8,25 +8,30 @@
...
     "css:prefix": "postcss --config postcss.config.js --replace \"dist/*.css\" \"!dist/*.min.css\"",
     "css:minify": "cleancss --format breaksWith=lf --source-map-inline-sources --batch --batch-suffix \".min\" \"dist/*.css\" \"!dist/*.min.css\"",
-    "test": "yarn css:lint",
-    "build:js": "node scripts/build.js",
-    "build": "yarn build:js && yarn build:css",
-    "dev": "node scripts/serve.js",
+    "css:lint": "stylelint --config .stylelintrc \"src/**/*.scss\" --cache --cache-location .cache/.stylelintcache",
+    "prejs": "yarn js:lint",
+    "js": "yarn js:compile",
+    "js:lint": "echo 'To be defined'",
```
adamkudrna, line 19: OK 🙂

package.json
```diff
@@ -8,25 +8,30 @@
...
+    "js:compile": "node scripts/build.js",
+    "lint": "npm-run-all --serial css:lint lint:commit",
+    "lint:commit": "yarn commitlint --from $(git describe --always --first-parent) --verbose",
+    "lint:commit:last": "yarn commitlint --from HEAD~1 --to HEAD --verbose",
```
adamkudrna, line 23: 👍🏻

adamkudrna (approved):

PR #12, opened 2021-10-20
Feat: Add translations for cs, en, sk and pl language #CCM-10

OndraM:


src/languages/cs.js
```diff
@@ -1,23 +1,19 @@
 export const config = {
   consent_modal: {
-    title: 'Tyto stránky používají cookies',
+    title: 'Tyto stránky využívají cookies',
     description:
-      'Cookies používáme, abychom zajistili správné fungování a bezpečnost našich stránek, tím pádem co nejlepší zkušenost při návštěvě. Kliknutím na „Přijmout všechny“ dáváte svůj souhlas s použitím cookies pro účely reklamy a analytik. Svá nastavení cookies můžete později kdykoliv změnit. Více informací naleznete v <a href="#">Prohlášení o cookies.</a>',
+      'Kliknutím na „Přijmout vše“ dáváte souhlas společnosti LMC s.r.o. k využívání souborů Cookies a dalších identifikátorů ve vašem zařízení. Použití těchto Cookies a dalších identifikátorů usnadní navigaci na stránkách, zobrazení personalizovaného obsahu, cílený marketing, analýzu využívání našich produktů a služeb. ' +
```
literat, line 5: Why have you used string concatenation here? If you just need to add line break you can use [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) instead.

```javascript
console.log(`string text line 1
string text line 2`);
// "string text line 1
// string text line 2"
```

literat (changes requested):

src/languages/cs.js
literat: Much better :-)

PR #16, opened 2021-10-22
Chore: Introduce publish package pipeline #CCM-8

literat:
* based on release publishing

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,38 @@
+name: Publish
+
+on:
+  release:
+    types:
+      - published
+
+jobs:
+  publish:
+    name: Publish package
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: 14
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+
+      - name: Build
+        run: yarn build
+
+      - name: Bump version
+        run: yarn version --`./bin/ci/semver.sh`
+
+      - name: Authenticate npm
+        run: ./bin/ci/npm-auth.sh
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+      - name: Publish
+        run: cd ./dist && yarn publish
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,12 +1,11 @@
 name: Publish
 
 on:
-  release:
-    types:
-      - published
+  create:
 
 jobs:
   publish:
+    if: ${{ startsWith(github.ref, 'refs/tags/v') }}
     name: Publish package
     runs-on: ubuntu-20.04
     steps:
@@ -23,9 +22,6 @@ jobs:
 
       - name: Build
         run: yarn build
-
-      - name: Bump version
-        run: yarn version --`./bin/ci/semver.sh`
 
       - name: Authenticate npm
         run: ./bin/ci/npm-auth.sh
--- a/bin/ci/semver.sh
+++ b/bin/ci/semver.sh
@@ -1,14 +1,14 @@
 #!/usr/bin/env bash
 
 function grep_word_from_git_commits_since_last_tag() {
-    local commits=$(git log --oneline $(git describe --tags --abbrev=0 @^)..@ --grep="$1" 2> /dev/null)
+    local commits
+    commits=$(git log --oneline $(git describe --tags --abbrev=0 @^)..@ --grep="$1" 2> /dev/null)
 
     # Check exit code of last command, eg. the one that gets the commits
     # and normalize exit status if there is some unpredictable error (git or grep fail)
     if [ $? -ne 0 ]; then
         echo 2
-    elif [ -z "$commits" ];
-    then
+    elif [ -z "$commits" ]; then
         echo 0
     else
         echo 1
@@ -16,10 +16,10 @@ function grep_word_from_git_commits_since_last_tag() {
 }
 
 function get_version() {
-    if [ `grep_word_from_git_commits_since_last_tag "BREAKING CHANGE"` -gt 0 ]
+    if [ "$(grep_word_from_git_commits_since_last_tag "BREAKING CHANGE")" -gt 0 ]
     then
         echo "major"
-    elif [ `grep_word_from_git_commits_since_last_tag "Feat"` -gt 0 ]
+    elif [ "$(grep_word_from_git_commits_since_last_tag "Feat")" -gt 0 ]
     then
         echo "minor"
     else
--- a/bin/ci/semver.sh
+++ b/bin/ci/semver.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+
+function grep_word_from_git_commits_since_last_tag() {
+    local commits=$(git log --oneline $(git describe --tags --abbrev=0 @^)..@ --grep="$1" 2> /dev/null)
+
+    # Check exit code of last command, eg. the one that gets the commits
+    # and normalize exit status if there is some unpredictable error (git or grep fail)
+    if [ $? -ne 0 ]; then
+        echo 2
+    elif [ -z "$commits" ];
+    then
+        echo 0
+    else
+        echo 1
+    fi
+}
+
+function get_version() {
+    if [ `grep_word_from_git_commits_since_last_tag "BREAKING CHANGE"` -gt 0 ]
+    then
+        echo "major"
+    elif [ `grep_word_from_git_commits_since_last_tag "Feat"` -gt 0 ]
+    then
+        echo "minor"
+    else
+        echo "patch"
+    fi
+}
+
+get_version
--- a/package.json
+++ b/package.json
@@ -35,7 +35,9 @@
     "prepare": "husky install",
     "changelog": "yarn changelog:from-tag",
     "changelog:from-tag": "conventional-changelog -p @lmc-eu/conventional-changelog-lmc-bitbucket -i CHANGELOG.md -s",
-    "changelog:from-beginning": "conventional-changelog -p @lmc-eu/conventional-changelog-lmc-bitbucket -i CHANGELOG.md -s -r 0"
+    "changelog:from-beginning": "conventional-changelog -p @lmc-eu/conventional-changelog-lmc-bitbucket -i CHANGELOG.md -s -r 0",
+    "version": "yarn build && git status && git add --all && git status",
+    "postversion": "git push --set-upstream origin main && git push --tags"
   },
   "dependencies": {
     "@lmc-eu/spirit-design-tokens": "^0.2.0",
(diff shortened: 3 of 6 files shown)


literat:
@OndraM Updated work flow to trigger it localy with `yarn release`, than check it and push. After that publish pipeline will be running by gh actions.

package.json
```diff
@@ -30,7 +30,11 @@
     "prepare": "husky install",
     "lint:commit:test": "yarn commitlint --from HEAD~1 --to HEAD --verbose",
     "changelog": "conventional-changelog -p @lmc-eu/conventional-changelog-lmc-bitbucket -i CHANGELOG.md -s",
-    "changelog:origin": "conventional-changelog -p @lmc-eu/conventional-changelog-lmc-bitbucket -i CHANGELOG.md -s -r 0"
+    "changelog:origin": "conventional-changelog -p @lmc-eu/conventional-changelog-lmc-bitbucket -i CHANGELOG.md -s -r 0",
+    "preversion": "yarn build",
+    "version": "git status && git add CHANGELOG.md package.json ./dist && git status",
```
OndraM, line 35: So if it re-uses the name of native [`yarn version`](https://classic.yarnpkg.com/en/docs/cli/version), it is run as part of it? Is it usual to override names of the native scripts like this? It confused me on first look 👀

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+
+      - name: Build
+        run: yarn build
+
+      - name: Authenticate npm
+        run: ./bin/ci/npm-auth.sh
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
OndraM, line 29: Why is this step needed? Isn't the following `yarn publish` enough? (It also receives `NPM_TOKEN`, so it is a bit confusing why it is needed for two steps).

package.json
OndraM: BTW isn't the changelog generation missing from the script? There is `git add CHANGELOG.md`, though none changelog is generated when running `yarn release`.

bin/ci/semver.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+function grep_word_from_git_commits_since_last_tag() {
+    local commits=$(git log --oneline $(git describe --tags --abbrev=0 @^)..@ --grep="$1" 2> /dev/null)
```
OndraM, line 4:
```suggestion
    local commits
    commits=$(git log --oneline "$(git describe --tags --abbrev=0 @^)"..@ --grep="$1" 2> /dev/null)
```

https://github.com/koalaman/shellcheck/wiki/SC2155#problematic-code-in-the-case-of-local

bin/ci/semver.sh
```diff
@@ -0,0 +1,28 @@
...
+
+    if [ $? -eq 128 ]; then
```
OndraM, line 6: Could you please add comment what does comparing with 128 mean? It looks a bit cryptic :)

bin/ci/semver.sh
```diff
@@ -0,0 +1,28 @@
...
+        echo 2
+    elif [ -z "$commits" ];
+    then
```
OndraM, line 9:
```suggestion
    elif [ -z "$commits" ]; then
```

bin/ci/semver.sh
```diff
@@ -0,0 +1,28 @@
...
+        echo 1
+    fi
+}
+
+function get_version() {
+    if [ `grep_word_from_git_commits_since_last_tag "BREAKING CHANGE"` -gt 0 ]
+    then
+        echo "major"
+    elif [ `grep_word_from_git_commits_since_last_tag "Feat"` -gt 0 ]
+    then
```
OndraM, line 21:
```suggestion
    elif [ $(grep_word_from_git_commits_since_last_tag "Feat") -gt 0 ]; then
```

bin/ci/semver.sh
```diff
@@ -0,0 +1,28 @@
...
+        echo 0
+    else
+        echo 1
+    fi
+}
+
+function get_version() {
+    if [ `grep_word_from_git_commits_since_last_tag "BREAKING CHANGE"` -gt 0 ]
+    then
```
OndraM, line 18:
```suggestion
    if [ "$(grep_word_from_git_commits_since_last_tag "BREAKING CHANGE")" -gt 0 ]; then
```
https://github.com/koalaman/shellcheck/wiki/SC2006
https://github.com/koalaman/shellcheck/wiki/SC2046

OndraM (changes requested):
It looks nice 👍

Could you also please document the release workflow into README? Maybe add new "Create new release"  section as part of [Development and contributing](https://github.com/lmc-eu/cookie-consent-manager/#development-and-contributing) section.

package.json
literat: Yes, the changelog command is missing. Thanks for notice.

literat: What do you mean with "re-uses" of the name? `pre` and `post` are standard prefixes and postfixes for running scripts before and after main script (https://docs.npmjs.com/cli/v7/using-npm/scripts#pre--post-scripts).

bin/ci/semver.sh
literat: Comment added.

bin/ci/semver.sh
literat: Fixed, thanks.

package.json
OndraM: I meant the `version` script itself, not the pre/post ones. Because there is already the native `yarn version` and here you are defining your own as an addition.

literat: No, this is same as `npm version`, see https://classic.yarnpkg.com/en/docs/cli/version. To get the version of yarn itself you must use `yarn --version`.

And of course with version command you can specify what should be done before committing and tagging (https://docs.npmjs.com/cli/v7/commands/npm-version#description).

bin/ci/semver.sh
OndraM: I though the "meaning" of this condition - the reason, why it is there. When does this condition happen in this scenario? Why is this some specific condition for exit code 128? Why not compare with other error codes? So some comment explaining this, because from reading the script I don't understand why is there such condition.

literat:
> It looks nice 👍
> 
> Could you also please document the release workflow into README? Maybe add new "Create new release" section as part of [Development and contributing](https://github.com/lmc-eu/cookie-consent-manager/#development-and-contributing) section.

Should I use CONTRIBUTING.md instead (https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors)? And add that part from readme to it as well?

package.json
OndraM: I didn't meant `--version` :) . But the npm docs provides an answer:

> Run the `version` script. These scripts have access to the new version in package.json (so they can incorporate it into file headers in generated files for example). Again, scripts should explicitly add generated files to the commit using git add.

This is something not explained in yarn documentation, even in the examples there is just preversion/postversion. But if its the same as in npm, I understand its purpose,  thanks.

literat: Ok, now I understand your confusion :-)

OndraM:
> Should I use CONTRIBUTING.md instead (https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors)? And add that part from readme to it as well?

I would now not bother with CONTRIBUTING.md for this project - we don't expect much external contributions... For now I think just short mention in readme would be enough. We can split it eventually later, if the readme become too long... (also this could be done in different PR, to not further extend scope of this one).

package.json
```diff
@@ -35,7 +35,11 @@
     "prepare": "husky install",
     "changelog": "yarn changelog:from-tag",
     "changelog:from-tag": "conventional-changelog -p @lmc-eu/conventional-changelog-lmc-bitbucket -i CHANGELOG.md -s",
-    "changelog:from-beginning": "conventional-changelog -p @lmc-eu/conventional-changelog-lmc-bitbucket -i CHANGELOG.md -s -r 0"
+    "changelog:from-beginning": "conventional-changelog -p @lmc-eu/conventional-changelog-lmc-bitbucket -i CHANGELOG.md -s -r 0",
+    "preversion": "yarn build",
+    "version": "yarn changelog && git status && git add CHANGELOG.md package.json ./dist && git status",
```
OndraM, line 40: Also, the `dist/` files build in preversion should be added.

OndraM (changes requested):

package.json
literat: They are there with `./dist` I presume.

bin/ci/semver.sh
literat: This is check for unpredictable behaviour. There are multiple commands in one and every single one could cause an error. So I have normalized the error status code to 2 when some unpredictable situation happend.

.github/workflows/publish.yaml
literat: `npm-auth` is creating `.npmrc` file with authentification configuration, sets right token and right registry. We can also omit this script and use `actions/setup-node@v2` with some configuration to do that for us (https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry).

But I want do that myself and also check the user with `npm whoami ` (https://docs.npmjs.com/cli/v7/commands/npm-whoami).

For publishing you need the token as well. So that is the reason why it is twice here.

Should I omit that script or add some comments to explain that?

package.json
OndraM: But they were not added to the repo as part of the last release: https://github.com/lmc-eu/cookie-consent-manager/tree/main/dist

literat: I do not understand what is wrong here. Could you be more specific, please?

OndraM: I thought that as part of the release, new version of dist files should be build and added to the repository.

However - you tagged the 0.2.0 version, but the [files in dist](https://github.com/lmc-eu/cookie-consent-manager/tree/main/dist) are unchanged.

literat: Oh, now I see the problem. My bad, I am sorry. However in the package itself they are modified. So this is problem only for examples of the 0.2.0 version. This release should fix it.

OndraM: Ok, we will see.

EYEgxfLXsAAItE9

bin/ci/semver.sh
OndraM: So if this is not specific to exit code 128 but generally "command caused an error", why not to check whether return code does not equal 0?

literat: So what do you suggest? Because this is working script and we spend a lot of time on this single detail. Maybe simple
```bash
if [ -z "$commits" ]; then
        echo 0
    else
        echo 1
    fi
```

is enough?

OndraM: Well if the purpose of the condition is to check whether the git command has not failed, then:

```diff
-     if [ $? -eq 128 ]; then 
+     if [ $? -ne 0 ]; then  
```

I thought there was some hidden meaning why check specifically the128 exit code, thus my confusion.

literat: Ok, fixed.

.github/workflows/publish.yaml
OndraM: No, this is fine, thanks for explanation 👍

OndraM (approved):
LGTM! :ship: 

(Also, could you please do a new release 0.3.0 right after this is merged?)

PR #18, opened 2021-10-22
Chore/release drafter

literat:


adamkudrna (approved):

literat:
I've updated this PR and remove release-drafter because we will be using semantic-release package in near future which can automagically create github release from conventional commits.

literat:
Only pr-labelling stays because I think it is a cool feature :-)

OndraM (approved):

PR #22, opened 2021-10-29
Update README, extend examples

OndraM:
Mainly README rearrangement. Further examples will follow.

I also removed the `autoload.js` script, because for the near future we expect one would need to provide at least one configuration option (product name), so this autoload case will not be possible anyway.


README.md
```diff
@@ -4,107 +4,198 @@ Provide configurable cookie consent banner with predefined LMC defaults. The UI
...
+
+### Custom methods
+
+To execute custom code which depends on cookie consent use callbacks:
+
+```js
+// ...
+initLmcCookieConsentManager(
+  {
+    'onAcceptAll': (cookie, cookieConsent) => {
```
literat, line 98: Maybe `onAccept` as a default method would be better in example.

README.md
```diff
@@ -4,107 +4,198 @@ Provide configurable cookie consent banner with predefined LMC defaults. The UI
...
+This will make the development server accessible http://localhost:3000/ .
+
+* The library with init function is served on: http://localhost:3000/dist/init.js
+* Javascript module: http://localhost:3000/dist/LmcCookieConsentManager.js
+* CSS: http://localhost:3000/dist/LmcCookieConsentManager.css
+
+### Contributing
 
-You can initialize cookie consent manager with following parameters:
+Follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) specification for commit messages.
```
literat, line 186: This is little bit misleading. Default conventional commits are in sentence case without leading caps. You should also link our company rules: https://github.com/lmc-eu/code-quality-tools/blob/main/packages/commitlint-config/index.js

README.md
```diff
@@ -4,107 +4,198 @@ Provide configurable cookie consent banner with predefined LMC defaults. The UI
...
+
+```sh
+yarn test
+```
+
+To fix code style according to Prettier rules run:
+
+```sh
+yarn format:fix
+```
```
literat, line 198: Maybe you should consider to add something like github actions badge with link to our actions ci which provide additional layer of code linting and testing.

literat (approved):

README.md
OndraM: For this scenario I don't think so - `onAccept` gets triggered even for "reject" aka "accept only necessary". So for case like this example, where you want to verify whether `functionality` category is allowed, you can use the more specific `onAcceptAll`.

README.md
OndraM: Ok, I will add link to this, however the issue is there is so far no written documentation for the LMC rules (except this configuration script). This is something that should be definitely added, so that people can learn and follow the rules.

README.md
OndraM: Ok, good idea. 👍 I will do it in another PR, or do you want to prepare it?

README.md
OndraM: Link to LMC rules added

README.md
literat: Ok, I will do it.

README.md
literat: Ok.

literat (approved):

PR #23, opened 2021-10-30
Feat: Enable management of third-party scripts 

OndraM:
This adds default configuration to enable third-party script management:
```js
page_scripts: true, // Manage third-party scripts loaded using <script>
```

If someone won't use this feature, the overhead in the vanilla library is one call of `var scripts = document.querySelectorAll('script[' + _config.script_selector + ']');` on init. I guess this is acceptable performance tradeoff?

Other option would be leave this as default and add this the configurable parameters. What do you think?

### DIFF
--- a/src/LmcCookieConsentManager.js
+++ b/src/LmcCookieConsentManager.js
@@ -51,16 +51,16 @@ const LmcCookieConsentManager = (args) => {
   const isFirstTimeAccept = !cookieConsent.validCookie(cookieName);
 
   cookieConsent.run({
-    current_lang: defaultLang,
-    auto_language: autodetectLang,
-    theme_css: themeCss,
-    cookie_name: cookieName,
-    cookie_expiration: 365,
-    use_rfc_cookie: true,
-    autorun: true,
-    delay: 0,
-    force_consent: false,
+    auto_language: autodetectLang, // Enable detection from navigator.language
+    autorun: true, // Show the cookie consent banner as soon as possible
+    cookie_expiration: 365, // 1 year
+    cookie_name: cookieName, // Predefined cookie name. Do not override.
+    current_lang: defaultLang, // Default language used when auto_language is false (or when autodetect failed)
+    delay: 0, // Show the modal immediately after init
+    force_consent: false, // Do not force the consent before page could be used
     hide_from_bots: true, // To be hidden also from Selenium
+    theme_css: themeCss, // Path to external CSS loaded by the component. Empty to disable.
+    use_rfc_cookie: true, // Store cookie content in RFC compatible format.
     gui_options: {
       consent_modal: {
         layout: 'bar', // box/cloud/bar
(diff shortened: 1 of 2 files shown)


literat:
> This adds default configuration to enable third-party script management:
> 
> ```js
> page_scripts: true, // Manage third-party scripts loaded using <script>
> ```
> 
> If someone won't use this feature, the overhead in the vanilla library is one call of `var scripts = document.querySelectorAll('script[' + _config.script_selector + ']');` on init. I guess this is acceptable performance tradeoff?
> 
> Other option would be leave this as default and add this the configurable parameters. What do you think?

I think that this is not as simple as you mean it. As I read in https://github.com/orestbida/cookieconsent#manage-third-party-scripts you must first update your script tags to `text/plain` with `data-` attribute. 

See https://github.com/orestbida/cookieconsent/blob/master/src/cookieconsent.js#L31 and https://github.com/orestbida/cookieconsent/blob/master/src/cookieconsent.js#L1261

So basically setting this option only to `true` will not work.

src/LmcCookieConsentManager.js
```diff
@@ -43,21 +43,21 @@ const LmcCookieConsentManager = (args) => {
...
-    force_consent: false,
+    auto_language: true, // Enable detection from navigator.language
+    autorun: true, // Show the cookie consent banner as soon as possible
+    cookie_expiration: 365, // 1 year
+    cookie_name: cookieName, // Predefined cookie name. Do not override.
+    current_lang: currentLang, // Default language used when auto_language is false (or when autodetect failed)
+    delay: 0, // Show the modal immediately after init
+    force_consent: false, // Do not force the consent before page could be used
     hide_from_bots: true, // To be hidden also from Selenium
+    page_scripts: true, // Manage third-party scripts loaded using <script>
```
literat, line 58: Using only this option without modification of script tags will not work, see https://github.com/orestbida/cookieconsent#manage-third-party-scripts

literat (changes requested):

src/LmcCookieConsentManager.js
OndraM: Sure, this will be documented in the readme, see #22: https://github.com/lmc-eu/cookie-consent-manager/pull/22/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R108

However without this option enabled it will not work even when product modifies the script tags. So the point is to make it enabled by default, and if the product decides to use this behavior, all they need to do is modify the script tags, as explained in the readme.

literat: I think that this is still misleading. If product want to manage third party cookie by cookie consent manager it must prepare those changes anyway. So it can also add another option to cookie consent manager.

You also wrote that product must provide at least one option in near future - https://github.com/lmc-eu/cookie-consent-manager/pull/22#issue-1039498802. So adding another one when it will want to manage third party cookie and also must update its script tags is not so much work.

Another handicap is that you are saying something like "hey, this options is on, so it will do something" but until you update script tags it will do nothing.

So my opinion is that we should leave it as is (false value) and the product developers must do those steps deliberatilly to make it work.

literat (approved):

src/LmcCookieConsentManager.js
OndraM: For now solved by adding documentation on how to disable the feature. Could be revisited lated.

PR #24, opened 2021-10-30
Configurable language autodetection

OndraM:
From talks with some engineers it seems the autodetection will be often disabled (they will just force the language based on current language version of the site). (Sidenote - maybe we should consider whether the autodetection should not be disabled by default...)

Furthermore, I renamed `currentLang` to `defaultLang`, which in my opinion better describes what the value actually means from configuration perspective.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ Optional config parameters could be provided on plugin initialization in the con
 ```js
 initLmcCookieConsentManager(
   {
-    'currentLang': 'cs',
+    'defaultLang': 'cs',
     'onAcceptAll': (cookie, cookieConsent) => {
       // custom code
     },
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -135,7 +136,8 @@ initLmcCookieConsentManager(
 
 | Option        | Type     | Default value                  | Description                                                                                                                               |
 |---------------|----------|--------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------|
-| `currentLang` | string   | 'cs'                           | Default language. One of `cs`, `en`, `sk`, `pl`. This language will be used when autodetect is disabled or when it fails.                 |
+| `defaultLang` | string   | 'cs'                           | Default language. One of `cs`, `en`, `sk`, `pl`. This language will be used when autodetect is disabled or when it fails.                 |
+| `autodetectLang`| string | true                           | Autodetect language from the browser. If autodetect fails or detects not supported language, fallback to `defaultLang`.<br>When disabled, force language to `defaultLang`. |
```
literat, line 140: `...or if unsupported language is detected...` seems better to me.

literat (approved):

README.md
OndraM: Changed

PR #29, opened 2021-11-02
Deps: Bump all dependencies

OndraM:


literat:
I think that the commit message should be `Deps: Bump vanilla-cookieconsent from 2.6.0 to 2.6.1` because you have updated only this particular dependency not all of them.

literat (approved):

OndraM:
> I think that the commit message should be `Deps: Bump vanilla-cookieconsent from 2.6.0 to 2.6.1` because you have updated only this particular dependency not all of them.

I'm not sure whats the convention in JS projects, but I not only added the version constrain to 2.6.1, I also ran `yarn upgrade` to upgrade all dependencies in the lock file. If I didn't change the package.json and only upgrade dependencies, what would be the commit message then?

literat:
Yes, you are right of course. I am used to use some kind of dependency automation like dependabot or renovate which suggests to use version pinning. After that are dependencies updated one by one and version change in `package.json` correlates with update of `yarn.lock`. So I am not doing "update all" anyway.

Maybe we should consider add this kind of automation in future.

literat (approved):

OndraM:
> Maybe we should consider add this kind of automation in future.

Yup, agreed, we can add depdendabot or something. Fell free to set it up, or we can do in issue for this.

PR #30, opened 2021-11-02
Feat: Introduce new design connected to Spirit design tokens #CCM-7

adamkudrna:
obrazek
obrazek
obrazek
obrazek
obrazek
obrazek

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,13 @@ The package is a wrapper around [Cookie Consent] by [Orest Bida].
 
 ## Basic usage
 
+Load default CSS in your `<head>`:
+
+```html
+<link rel="preconnect" href="https://cdn.jsdelivr.net">
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@lmc-eu/cookie-consent-manager@0.3.1/LmcCookieConsentManager.min.css">
+```
+
 Load the script and initialize the plugin right before ending `</body>` tag:
 
 ```html
@@ -22,6 +29,24 @@ window.addEventListener('load', function () {
 
 This will load the plugin from CDN and initialize the plugin with default settings. See [examples/index.html](examples/index.html).
 
+## Use default web font, or not?
+
+If you are going to use the plugin with the default theme, you may also want to
+load default Spirit web font which is used by the plugin:
+
+```html
+<link rel="preconnect" href="https://fonts.googleapis.com">
+<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Inter:wght@400;700&display=swap">
+```
+
+If Inter font is not provided (or installed in user's system), all texts in
+cookie consent UI default to whatever current `font-family` is set to. As you
+cannot predict what fonts are available on user's side, and because this
+behavior may change in future versions of Spirit Design Tokens, we encourage you
+either to load the default web font as shown above, or to explicitly specify
+the desired font yourself (head to [Theming](#theming) to see how).
+
 ## Loading the plugin
 
 ### Via CDN or static file
@@ -29,6 +54,7 @@ This will load the plugin from CDN and initialize the plugin with default settin
 You can load the plugin from a CDN, as in the basic example above.
 
 ```html
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@lmc-eu/cookie-consent-manager@0.3.1/LmcCookieConsentManager.min.css">
 <script defer src="https://cdn.jsdelivr.net/npm/@lmc-eu/cookie-consent-manager@0.3.1/init.js"></script>
 ```
 
@@ -50,7 +76,7 @@ via npm package [@lmc-eu/cookie-consent-manager](https://www.npmjs.com/package/@
     ```
     or
     ```sh
-    npm i @lmc-eu/cookie-consent-manager
+    npm install --save @lmc-eu/cookie-consent-manager
     ```
 
 2. Import the module in your javascript:
@@ -62,14 +88,17 @@ via npm package [@lmc-eu/cookie-consent-manager](https://www.npmjs.com/package/@
     });
     ```
 
-3. Include default styles:
-    ```scss
-    @use '../node_modules/@lmc-eu/cookie-consent-manager/LmcCookieConsentManager';
-
-    :root {
-        // override default config
-    }
-    ```
+3. Include default CSS in your HTML:
+   ```html
+   <link rel="stylesheet" href="node_modules/@lmc-eu/cookie-consent-manager/LmcCookieConsentManager.min.css">
+   ```
+   or in your Sass stylesheet:
+   ```scss
+   @use 'node_modules/@lmc-eu/cookie-consent-manager/LmcCookieConsentManager.css';
+   ```
+   Please mind the `.css` extension used in the Sass example. Using the provided
+   `.scss` stylesheet is only [recommended](#with-spirit-design-system) for
+   projects that are built with [Spirit Design System].
 
 ## Manage features depending on given consent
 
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -25,7 +25,7 @@
     "lint": "npm-run-all --serial css:lint lint:commit",
     "lint:commit": "yarn commitlint --from $(git describe --always --first-parent) --verbose",
     "lint:commit:last": "yarn commitlint --from HEAD~1 --to HEAD --verbose",
-    "prebuild": "rm -rf dist && mkdir -p dist && cp package.json README.md dist/",
+    "prebuild": "rm -rf dist && mkdir -p dist/scss && cp package.json README.md src/LmcCookieConsentManager.scss dist/ && cp -r src/scss/* dist/scss/",
```
literat, line 28: This is going to be so long. May I suggest to create some new scripts?

```json
"build:clean": "rm -rf dist && mkdir -p dist/scss",
"build:copy": "cp package.json README.md src/LmcCookieConsentManager.scss dist/ && cp -r src/scss/* dist/scss/"
"prebuild": "npm-run-all --serial build:clean build:copy"
""
```

literat (changes requested):

literat (approved):

crishpeen (approved):

README.md
```diff
@@ -9,10 +9,17 @@ The package is a wrapper around [Cookie Consent] by [Orest Bida].
 
 ## Basic usage
 
+Load default CSS in your `<head>`:
+
+```html
+<link rel="preconnect" href="https://cdn.jsdelivr.net" />
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@lmc-eu/cookie-consent-manager@v0.2.0/LmcCookieConsentManager.min.css" />
+```
```
OndraM, line 17: BTW this may be subject of change - we can by default load the css through the theme_css feature.

Lets discuss this css/spirit-related issues on Thursday.

README.md
```diff
@@ -181,6 +202,106 @@ Each configured callback receives two params:
...
+// close to the end of your stylesheet as it contains CSS selectors with high
+// specificity.
+@use '@lmc-eu/cookie-consent-manager/LmcCookieConsentManager';
+```
+
+And set up a [Sass load path] so the Sass compiler can find stylesheets located
+in the `node_modules` directory (you will already have the path to your design
+tokens there as required by [Spirit Web]):
+
+```bash
```
OndraM, line 226:
```suggestion
```sh
```

(for consistency)

examples/index.html
```diff
@@ -1,17 +1,27 @@
...
 <html lang="en">
 <head>
-  <meta charset="UTF-8">
+  <meta charset="UTF-8" />
+  <link rel="preconnect" href="https://fonts.googleapis.com" />
+  <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+
   <title>Cookie Consent Manager</title>
-  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta name="viewport" content="width=device-width, initial-scale=1" />
```
OndraM, line 9:
```suggestion
  <meta name="viewport" content="width=device-width, initial-scale=1">
```

> As it is a void element, the start tag must be present and the **end tag must not be present**.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta


(also for the other link/meta tags below and above)

OndraM (changes requested):

package.json
```diff
@@ -38,7 +40,7 @@
     "changelog:from-beginning": "conventional-changelog -p @lmc-eu/conventional-changelog-lmc-bitbucket -i CHANGELOG.md -s -r 0"
   },
   "dependencies": {
-    "@lmc-eu/spirit-design-tokens": "^0.2.0",
+    "@lmc-eu/spirit-design-tokens": "~0.4.0",
```
OndraM, line 43:
```suggestion
    "@lmc-eu/spirit-design-tokens": "^0.4.0",
```

https://developer.happyr.com/always-use-caret-instead-of-tilde

examples/index.html
adamkudrna: > 6. Then, if the element is one of the void elements, or if the element is a foreign element, then there **may be a single U+002F SOLIDUS character (/)**. This character has no effect on void elements, but on foreign elements it marks the start tag as self-closing.

https://html.spec.whatwg.org/multipage/syntax.html#start-tags

🤔

package.json
adamkudrna: I disagree. `0.x` versions mark an [early stage of development](https://semver.org/#spec-item-4) with breaking changes happening even during minor version bumps. Allowing upgrade to a higher minor version of `0.x` thus may break the app.

Once the referred package version is `>=1`, then it's time to change `~` to `^`.

examples/index.html
OndraM: > This character has no effect on void elements

Soo, why should we have something which has no effect? :)  Both is valid, thats true, but for consistency - we use `<br>`, `<hr>` etc. in html5, so why this should be an exception?

package.json
OndraM: > Allowing upgrade to a higher minor version of 0.x thus may break the app.

But `^0.4.0` does not allow this. It means `>=0.4.0  < 0.5.0`, so "do safe upgrades without BC". So in the pre-1.x version, the behavior is effectively the same as using tilde. However after 1.x, the caret `^` follows safe semver upgrades (unlike tilde, which behaves differently and `~1.2.3` won't upgrade to `1.3.0`) .

So exactly what you wrote is the reason why use caret, for both 0.x and 1.x+ versions, without a need to remember you should change it after 1.x.

https://github.com/npm/node-semver#caret-ranges-123-025-004
https://semver.npmjs.com/

literat: @OndraM @adamkudrna And what about to pin the version to exact match, eg. `"@lmc-eu/spirit-design-tokens": "0.4.0",` and let dependabot to take care of updates?

Everybody got the same version and breaking and other changes would be under control in separate PRs. And also there will be no arguing about tilda or caret ;-)

examples/index.html
adamkudrna: I've been always used to writing trailing `/` everywhere, `<br />` or `<hr />` being no exception. In React projects, ESlint gives me no other choice anyway.

Generally speaking, when there are more possible ways of syntax in a code, I prefer the one which is easier to read and understand, even if it means more writing. To me, trailing slash denotes a self-closing element, as opposed to elements with starting and ending tags, and I'm able to recognize it even without reading actual tag names.

package.json
adamkudrna: > So in the pre-1.x version, the behavior is effectively the same as using tilde.

Nice one, didn't know that!

examples/index.html
OndraM: Understand your point and read it carefully, but still don't agree its valid for this (eg. this is not XML like JSX). So I don't think there is any benefit in doing thing differently than what is currently the de-facto standard / mostly used. From quick research, the no-slash version is also the usual convention in LMC products. So having example with different conventions will not make it easier for anyone.

So let move on with this - and go with the short version and no closing tags please :)

OndraM: (Or maybe just don't spend time on this now and let it be 🎶 - I will do some more cleanups in #37, so I can do it altogether...)

README.md
```diff
@@ -183,6 +212,111 @@ Each configured callback receives two params:
...
+properties as described above, this time scoped to the `.c_darkmode` class:
+
+```css
+.c_darkmode {
+  --lmcccm-bg: #000;
+}
+```
+
+Again, any customization through CSS custom properties must be done **after**
+the original plugin CSS has been loaded.
```
OndraM, line 318: This complicates the ease of use a bit... 🤯

OndraM (changes requested):
Remove all the `dist/` file changes pls :)

README.md
adamkudrna: You're fine when loading plugin's CSS via `<link>` in `<head>`. Things get a little bit more complicated with `themeCss` though — then you need to bundle the original CSS with your custom properties.

adamkudrna: Improved, see `fixup!`.

OndraM (approved):
LGTM, thanks! 👍

literat (approved):

PR #33, opened 2021-11-02
Chore: Automatically replace package version in readme file to actual

literat:
Simple regex: https://regex101.com/r/mufpT0/1/

### DIFF
--- a/scripts/readme-replace-version.js
+++ b/scripts/readme-replace-version.js
@@ -1,0 +1,15 @@
+const replace = require('replace-in-file');
+const packageJson = require('../package.json');
+const options = {
+  files: 'README.md',
+  from: /cookie-consent-manager@v([0-9]\.[0-9]\.[0-9])/gm,
+  to: `cookie-consent-manager@v${packageJson.version}`,
+};
+
+replace(options)
+  .then((results) => {
+    console.log('Replacement results:', results);
+  })
+  .catch((error) => {
+    console.error('Error occurred:', error);
+  });
--- a/scripts/readme-replace-version.js
+++ b/scripts/readme-replace-version.js
@@ -2,8 +2,8 @@ const replace = require('replace-in-file');
 const packageJson = require('../package.json');
 const options = {
   files: 'README.md',
-  from: /cookie-consent-manager@v([0-9]\.[0-9]\.[0-9])/gm,
-  to: `cookie-consent-manager@v${packageJson.version}`,
+  from: /cookie-consent-manager@([0-9]\.[0-9]\.[0-9])/gm,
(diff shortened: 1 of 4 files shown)


adamkudrna (approved):

scripts/readme-replace-version.js
```diff
@@ -0,0 +1,15 @@
+const replace = require('replace-in-file');
+const packageJson = require('../package.json');
+const options = {
+  files: 'README.md',
+  from: /@v([0-9].[0-9].[0-9])/gm,
+  to: `@v${packageJson.version}`,
```
OndraM, line 6:
```suggestion
  from: /(cookie-consent-manager@v)([0-9]\.[0-9]\.[0-9])/gm,
  to: `cookie-consent-manager@v${packageJson.version}`,
```

- `.` will match any character, thats probably not what we want
- I'd expand the matched part, to avoid any possible unintentional replacements.  Like if I write in readme: "Do not use v1.2.3, it is bad!", it will get replaced to current version :slightly_smiling_face:  . This is almost [real-world scenario](https://github.com/lmc-eu/cookie-consent-manager/pull/30/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R252).
- BTW is the `v`-prefix still [supposed to be there](https://github.com/lmc-eu/cookie-consent-manager/pull/36)?

OndraM (changes requested):
Maybe change the commit message: `actual` (skutečný, opravdový) -> `current` (současný, aktuální)

scripts/readme-replace-version.js
literat: Thank you for you suggestion. I've just applied it.

About that `v` prefix on jsdelivr I think that this is on us what we want to use. I've tried both urls with `v` and without `v` and they are working.

- https://cdn.jsdelivr.net/npm/@lmc-eu/cookie-consent-manager@v0.3.0/init.js
- https://cdn.jsdelivr.net/npm/@lmc-eu/cookie-consent-manager@0.3.0/init.js

literat: But I think because we are tagging without `v` so we should also use urls. without `v`.

OndraM: +1 for omitting the v :)

literat: `v` ommitet and commit message changed.

OndraM (approved):
If you have a chance, please change the commit message as part of rebase :)

literat:
> If you have a chance, please change the commit message as part of rebase :)

The commit message has been changed already.

PR #34, opened 2021-11-02
Chore/package keywords

literat:


adamkudrna (approved):

package.json
```diff
@@ -11,6 +17,15 @@
...
   },
+  "contributors": [
+    {
+      "author": "Ondřej Machulda <ondrej.machulda@lmc.eu>"
+    },{
+      "author": "Tomáš Litera <tomas.litera@lmc.eu>"
+    },{
+      "author": "Adam Kudrna <adam.kudrna@lmc.eu>"
+    }
+  ],
```
OndraM, line 28: Scheme validator complains: `Missing required property 'name' `

And it seems to be right: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#people-fields-author-contributors

BTW default json format in phpstrom/webstrom puts each `{` on a new line, so we may stick with this - or is there some other rule?

OndraM (changes requested):

package.json
literat: 🤔  In docs you mention I saw the shortened version and use it. So I am confused if it is valid or not.

literat: Fixed formatting. My formatter just won't run :-(

OndraM: The shortened version seems to be for a case when there is just one person (ie. the filed is not an array).

However this looks like a valid shorthand version for array of people:

```json
  "contributors": [
    "Ondřej Machulda <ondrej.machulda@lmc.eu>",
    "Tomáš Litera <tomas.litera@lmc.eu>",
    "Adam Kudrna <adam.kudrna@lmc.eu>"
  ],
```

:shrug:

literat: You just try that or read it somewhere else?

OndraM: String value of the array is allowed in package.json schema used by PhpStorm etc.. I also found it being used in some packages.

```json
"contributors": {
      "type": "array",
      "items": {
        "$ref": "#/definitions/person"
      }
    },
...
"person": {
      "type": ["object", "string"],
...
    },
```

However it is a grey zone :smile:, I don't know whether is is supposed to be like that. I'm a quite surprised there is no offical package.json validation schema... :flushed: . [Like for composer.json](https://github.com/composer/composer/blob/main/res/composer-schema.json). 🤷

literat: I have just found maybe [unofficial schema](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/package.json) for package.json. On npmjs.com there is no sign about package.json schema validation. So I mean it is json file anyway.

I doubt it is even possible to have schema for it because many packages like eslint, jest, babel and more can use package.json for storing their configuration. For example instead of prettierrc you have and object in package.json. Just see https://prettier.io/docs/en/configuration.html and other packages can be configured the same way.

I have also updated `contributors` to be simple array of string so this should work just fine as you suggested.

OndraM (approved):

package.json
OndraM: > I doubt it is even possible to have schema for it because many packages like eslint, jest, babel and more can use package.json for storing their configuration.

This sounds a bit spooky :jack_o_lantern: , if anyone can store anything anywhere, with no rules...

To compare with Composer - it has for these purposes `extra` field. Rest of the file is not a freestyle :shrug:

literat: Maybe it is time to write some rfc https://github.com/npm/rfcs/issues :-)

PR #35, opened 2021-11-02
Chore: Introduce dependabot configuration for automated deps updates

literat:


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+version: 2
```
OndraM, line 1: We can rename to dependabot.yaml (to use recommended `yaml` extension everywhere consistently) It should work, as per https://github.com/dependabot/feedback/issues/874

literat: Ok, file was renamed.

OndraM (approved):

PR #36, opened 2021-11-02
Chore: Set git tag version to plain semver string without prefix

literat:


OndraM (approved):
And what is the default prefix?

literat:
> And what is the default prefix btw?

Default prefix is set to `v` eg. `v0.1.0`

PR #37, opened 2021-11-02
Docs: Revamp examples #CCM-20

OndraM:
These are revamped examples - they were prepared for todays presentation, so not everything is as clean as it should. I welcome your suggestions - what examples add, whether the structure should be somehow changed etc.

It is based on bootstrap 5, as I needed something to make it quickly and good looking.

I set GitHub to deploy this branch here: https://lmc-eu.github.io/cookie-consent-manager/examples/

.. so that anyone can see the examples online right now, because the time is pressing us a bit.

~~(Note: the branch is targeted to `test/demo-integration`, which is temporary branch with current main + merged #30. I will switch it to main once #30 is merged).~~

### DIFF
--- a/examples/callbacks.html
+++ b/examples/callbacks.html
@@ -1,0 +1,117 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Cookie Consent Manager – callbacks example</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+
+    <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.1.3/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-1BmE4kWBq78iYhFldvKuhfTAU6auU8tT94WrHftjDbrCEXSU1oBoqyl2QvZ6jIW3" crossorigin="anonymous">
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap-icons@1.7.0/font/bootstrap-icons.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Inter:wght@400;700&display=swap">
+    <script src="./functions.js" defer></script>
+
+    <link rel="stylesheet" href="../dist/LmcCookieConsentManager.css"><!-- You will load this from CDN instead -->
+</head>
+<body>
+
+<header class="container mt-4">
+    <h1>LMC Cookie Consent Manager</h1>
+</header>
+
+<div class="container mt-4">
+    <ul class="nav nav-tabs">
+        <li class="nav-item">
+            <a class="nav-link" href=".">Basic example</a>
+        </li>
+        <li class="nav-item">
+            <a class="nav-link" href="./configuration.html">Extended configuration</a>
+        </li>
+        <li class="nav-item">
+            <a class="nav-link active" href="./callbacks.html">Callbacks</a>
+        </li>
+        <li class="nav-item">
+            <a class="nav-link" href="./theming.html">Theming</a>
+        </li>
+        <li class="nav-item">
+            <a class="nav-link" href="https://github.com/lmc-eu/cookie-consent/" target="_blank">Documentation <i class="bi bi-box-arrow-up-right"></i></a>
+        </li>
+    </ul>
+</div>
+
+<main class="container mt-4">
+    <div class="row">
+        <div class="col">
+            <p class="lead">
+                This is an example of configurable callbacks.
+            </p>
+
+            <p>
+                <a href="" class="btn btn-warning" onclick="removeCookieAndReload();">
+                    Remove <code>lmc_ccm</code> cookie and reload
+                </a>
+            </p>
+
+            <h3 class="mt-4">Callbacks</h3>
+
+            <div class="row">
+                <div class="col-md-6 col-xs-12">
+
+                <ul class="list-group">
+                    <li class="list-group-item" id="callback-onAccept">onAccept</li>
+                    <li class="list-group-item" id="callback-onAcceptAll">onAcceptAll</li>
+                    <li class="list-group-item" id="callback-onAcceptOnlyNecessary">onAcceptOnlyNecessary</li>
+                    <li class="list-group-item" id="callback-onFirstAccept">onFirstAccept</li>
+                    <li class="list-group-item" id="callback-onFirstAcceptAll">onFirstAcceptAll</li>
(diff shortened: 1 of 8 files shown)


examples/callbacks.html
```diff
@@ -0,0 +1,117 @@
...
+                </div>
+            </div>
+        </div>
+    </div>
+</main>
+
+<script defer src="../dist/init.js"></script>
+<script>
+  const markCallbackCalled = function (id) {
+    const element = document.getElementById('callback-' + id);
```
literat, line 78: If you are using `const` you can also use template literals for string concatenation, eg.
```
`callback-${id}`
```

examples/callbacks.html
```diff
@@ -0,0 +1,117 @@
...
+    element.classList.add('list-group-item-success');
+    element.innerText += ' – called'
+  }
+
+  window.addEventListener('load', function () {
+    window.ccm = initLmcCookieConsentManager( // note we store cookieConsent instance in global window.ccm variable
+      {
+        autodetectLang: false, // do not detect language from the browser
+        defaultLang: 'en', // use 'en' language by default
+        onAccept: (cookie, cookieConsent) => { // any type of consent detected
```
literat, line 88: I prefer to add comments above the function and not to an open bracket. Seems to me better for copy and formatting and reading the code (I see comment first and then function not opposite as here).

examples/callbacks.html
```diff
@@ -0,0 +1,117 @@
...
+
+<main class="container mt-4">
+    <div class="row">
+        <div class="col">
+            <p class="lead">
+                This is an example of configurable callbacks.
+            </p>
+
+            <p>
+                <a href="" class="btn btn-warning" onclick="document.cookie = 'lmc_ccm=; Max-Age=0; path=/; domain=' + location.hostname;">
```
literat, line 51: I think it would be better define this onClick as a named function and use it here. For better code readability I prefer something like `onclick="removeCookieAndReload()"`.

examples/configuration.html
```diff
@@ -0,0 +1,91 @@
...
+
+<main class="container mt-4">
+    <div class="row">
+        <div class="col">
+            <p class="lead">
+                This is an example of extended configuration.
+            </p>
+
+            <p>
+                <a href="" class="btn btn-warning" onclick="document.cookie = 'lmc_ccm=; Max-Age=0; path=/; domain=' + location.hostname;">
```
literat, line 51: This is the same as in `callbacks.html`. Please define a function or use external file where you store those functions. Keep it DRY.

examples/index.html
```diff
@@ -1,51 +1,131 @@
...
         },
-        'onFirstAcceptAll': () => {
-          console.log('All cookies were just accepted by the user');
-        }
       }
     );
   });
+
+  const markConsentCategoryAllowed = function (category) {
+    const element = document.getElementById('consent-' + category);
```
literat, line 125: You can use template literals.

literat (changes requested):

examples/configuration.html
OndraM: True, will do :) .This was done a bit in a hurry... :upside_down_face:

examples/callbacks.html
OndraM: This kinda reminded me PHP's curly braces variables in strings, which we are avoiding for like 10 years, all over again :upside_down_face:  . But if this is considered ok in ES, why not, I will change :)

examples/callbacks.html
OndraM: 🤔 I just tried, but then it looks weird if the other "scalar" parameters have the comment inline... And it would be unusual and long to have it on the line before for them as well, imho :shrug:

```js
      {
        autodetectLang: false, // do not detect language from the browser
        defaultLang: 'en', // use 'en' language by default
        config: { // override default config of the underlying library, see https://github.com/orestbida/cookieconsent#all-available-options
          delay: 500, // show cookie consent banner after 500ms
          page_scripts: false, // disable third-party script management, see https://github.com/lmc-eu/cookie-consent-manager#third-party-scripts-loaded-via-script
        },
      }
```
vs
```js
      {
        // do not detect language from the browser
        autodetectLang: false,
        // use 'en' language by default
        defaultLang: 'en',
        // override default config of the underlying library, see https://github.com/orestbida/cookieconsent#all-available-options
        config: { 
          // show cookie consent banner after 500ms
          delay: 500, 
          // disable third-party script management, see https://github.com/lmc-eu/cookie-consent-manager#third-party-scripts-loaded-via-script
          page_scripts: false,
        },
      }
```

What do you think?

examples/configuration.html
OndraM: Moved to external file, have a look :)

examples/index.html
OndraM: Done

examples/callbacks.html
literat: I've done quick research and clear to myself what visually bothers me. I have no problem with inline comments but those in curly braces are bad, I think.

Please see an eslint rule for that: https://eslint.org/docs/rules/no-inline-comments

So `autodetectLang: false, // do not detect language from the browser` is ok.

This is bad and comment should be above the function.
```
onAcceptAll: (cookie, cookieConsent) => { // consent for all cookies detected
          markCallbackCalled('onAcceptAll');
```

I am not sure about this
```
config: { // override default config of the underlying library, see https://github.com/orestbida/cookieconsent#all-available-options delay: 500, // show cookie consent banner after 500ms
```

I think that it is wrong and should be as like this
```
// override default config of the underlying library, see https://github.com/orestbida/cookieconsent#all-available-options
config: {
```

But I am not 100 % sure because it can be weird
```
      {
        autodetectLang: false, // do not detect language from the browser
        defaultLang: 'en', // use 'en' language by default
        // override default config of the underlying library, see https://github.com/orestbida/cookieconsent#all-available-options
        config: { 
          delay: 500, // show cookie consent banner after 500ms
          page_scripts: false,  // disable third-party script management, see https://github.com/lmc-eu/cookie-consent-manager#third-party-scripts-loaded-via-script
        },
      }

```

All three most used eslint configs (standard, recommended, airbnb) have this rule turned off (https://sqren.github.io/eslint-compare/). So maybe the problem is just in me :-)

literat (changes requested):

examples/callbacks.html
OndraM: Yes, I'm conflicted exactly about the last example, which triggers OCD :-D . I did a quick search in node_modules and comment after opening curly braces is used quite often. However this is nothing important to me, we can change it if you want - though I'd prefer if this would be part of some LMC-wide eslint rules which we would just obey here, so that we don't have to decide this here. 🤷

OndraM:
Rebased to `main` and ready for final review 🚀

examples/callbacks.html
literat: Ook, let it be as is.

literat (approved):

PR #40, opened 2021-11-04
Chore: Remove prepare script before publishing package

literat:
* `set-script` works from npm v7 and we are on node v14 and npm v6
  * we are removing prepare script with husky install because it is
    a lifecycle script that is runned before publish
  * we are publishing from ./dist directory where are no deps installed

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -21,7 +21,7 @@ jobs:
         run: yarn install --frozen-lockfile
 
       - name: Disable husky commit hooks
-        run: npm set-script prepare ""
+        run: yarn remove-prepare
 
       - name: Build
         run: yarn build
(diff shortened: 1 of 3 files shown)


literat:
Fixing my problem because I've used node v17 with npm v7 when developing this and didn't notice that we use node v14 in CI pipeline.

🤔 Maybe we should switch to [node v16 which is active LTS](https://nodejs.org/en/about/releases/).

adamkudrna (approved):

adamkudrna:
> 🤔 Maybe we should switch to [node v16 which is active LTS](https://nodejs.org/en/about/releases/).

Would it mean we could use `npm set-script` instead of manipulating `package.json`?

Also, I think shift to v16 should be started in all LMC projects then.

literat:
> Would it mean we could use `npm set-script` instead of manipulating `package.json`?

Yes, we can use `npm set-script` but it also manipulates `package.json`.

PR #53, opened 2021-11-09
Chore: Check for broken links in markdown

OndraM:


.github/markdown-link-check-config.json
```diff
@@ -0,0 +1,13 @@
+{
+  "ignorePatterns": [
+    {
+      "pattern": "^https://github.com/lmc-eu/cookie-consent-manager/(pull|issues|commit)/"
+    },
+    {
+      "pattern": "^https://github.com/projects/lmc-eu/repos/cookie-consent-manager/"
```
OndraM, line 7: This is because links in changelog are broken: https://github.com/lmc-eu/cookie-consent-manager/blob/main/CHANGELOG.md

This should be resolved in the future by using different/adjusted changelog generator. Because current links are made for our internal Bitbucket and JIRA, not github. // cc @literat

adamkudrna (commented):
How about JIRA links `#CCM-<number>`? They don't return 404, but they lead to Issues pane on GH which is empty. I think these links are useful for linking GH commits to JIRA issues but we need to rewrite them in the changelong too.

OndraM:
Yes, this needs the be resolved in the changelog generator as well... These links should lead to our internal JIRA but should be ignored here, because it is not reachable from the internet.

OndraM:
Ping @literat @adamkudrna @crishpeen - can we get this merged?

literat (approved):

PR #54, opened 2021-11-10
Introduce ES module entry point

literat:


literat:
Example in `examples/module` will be working with new version published. For simulating the current behaviour do these steps:
- run `yarn build`
- go to `dist`
- run `yarn pack` which creates `*.tgz` export of package
- go to `examples/module`
- install tgz using `yarn add <path-to-tgz>`
- run `yarn build`
- open `index.html` in your browser and you should see cookie consent

OndraM:
> This was little bit tricky because vanilla-cookie-consent was exported as main in package.json and webpack prioritize main over module.

Isn't this something which could be adjusted in the upstream vanilla package? Or they are doing it properly and the issue is purely ours?

OndraM (changes requested):
The change itself LGTM, thanks!

However - I don't think it worth it adding the webpack example in `examples/module/` and rearranging all the other examples into `examples/browser/` folder.
- It makes accessing basic examples a bit less obvious, as they are hidden in on more subfolder
- It includes code which will need constant maintenance (constant yarn.lock and package.json.updates, for instance), but the benefit is not clear
- It IMO does not include anything actually interesting worth showing as an example - it is is just using default webpack.config.js, the same include statement as in readme etc. Just purely basic stuff, which IMHO don't need an example... 
- So it is not obvious to me, example of what it actually showcases and why should someone use this example...

=> I think just an example of webpack integration (ie. extending the part "Via npm") in readme may be enough.

literat:
> > This was little bit tricky because vanilla-cookie-consent was exported as main in package.json and webpack prioritize main over module.
> 
> Isn't this something which could be adjusted in the upstream vanilla package? Or they are doing it properly and the issue is purely ours?

Well, using `main` in `package.json` for this type of package is not entirely wrong. As you can see also jQuery use it same way. Problem is that both packages are client-side package and they do not have default `module.export`, eg. they are not CommonJS (Node.js, server-side) packages. So they should technically use `browser` field in `package.json`, see https://docs.npmjs.com/cli/v7/configuring-npm/package-json#main

But problem with ESM export is purely ours.

literat:
> However - I don't think it worth it adding the webpack example in examples/module/ and rearranging all the other examples into examples/browser/ folder.

Well, now we have multiple examples and I think that it is fine.
 
> It makes accessing basic examples a bit less obvious, as they are hidden in on more subfolder

Ok, but what is "basic" example? Browser one or with import?

> It includes code which will need constant maintenance (constant yarn.lock and package.json.updates, for instance), but the benefit is not clear

Not constant, IMO, only when necessary. It is an example based on dist files.

> It IMO does not include anything actually interesting worth showing as an example - it is is just using default webpack.config.js, the same include statement as in readme etc. Just purely basic stuff, which IMHO don't need an example...

Well, I want to add a readme to those examples as well with some additional information which we can link form main readme. Also this basic stuff is also some kind of integration test, so we know that something is not broken.

> So it is not obvious to me, example of what it actually showcases and why should someone use this example...
> => I think just an example of webpack integration (ie. extending the part "Via npm") in readme may be enough.

If someone in company uses another tool we should add an example for that as well. Just for simple "It's working, dude!" ;-)

literat:
- [ ] Remove yarn.lock
- [ ] Remove bundle.js
- [ ] Add Readme
- [ ] Update root readme links
- [ ] Better directory structure

examples/webpack-with-cjs/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+Execute [npm][npm] or [Yarn][yarn] to bootstrap the example:
+
+```shell
+npm install
+# or
+yarn
+```
+
+Then run `build` script and open an `index.html` in your browser.
```
adamkudrna, line 15:
```suggestion
Then run `build` script and open `index.html` in your browser.
```

examples/webpack-with-cjs/index.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Cookie Consent Manager – module example</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+
+    <link rel="stylesheet" href="../../dist/LmcCookieConsentManager.min.css"><!-- You will load this from CDN instead -->
```
adamkudrna, line 8: I think we could add the default web font to all examples: https://github.com/lmc-eu/cookie-consent-manager#use-default-web-font-or-not

examples/webpack-with-esm/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+Execute [npm][npm] or [Yarn][yarn] to bootstrap the example:
+
+```shell
+npm install
+# or
+yarn
+```
+
+Then run `build` script and open an `index.html` in your browser.
```
adamkudrna, line 15:
```suggestion
Then run `build` script and open `index.html` in your browser.
```

examples/with-callbacks/README.md
```diff
@@ -0,0 +1,9 @@
+# Cookie Consent Manager with Callbacks
+
+This example shows how to use callbacks of Cookie Consent Manager.
+
+## How to use
+
+Open `callbacks.html` into your browser.
```
adamkudrna, line 7:
```suggestion
Open `callbacks.html` in your browser.
```

examples/with-extended-configuration/README.md
```diff
@@ -0,0 +1,9 @@
+# Cookie Consent Manager with extended configuration
+
+This example shows how to use extended configuration of Cookie Consent Manager.
+
+## How to use
+
+Open `configuration.html` into your browser.
```
adamkudrna, line 7:
```suggestion
Open `configuration.html` in your browser.
```

examples/with-theme/README.md
```diff
@@ -0,0 +1,9 @@
+# Cookie Consent Manager with theme
+
+This example shows how to use theme of Cookie Consent Manager.
+
+## How to use
+
+Open `theming.html` into your browser.
```
adamkudrna, line 7:
```suggestion
Open `theming.html` in your browser.
```

adamkudrna (approved):

OndraM:
The examples would be for me as a developer using this library still a bit confusing.

We basically want to present two types of examples:
1. configuration examples (how to adjust various configuration, including theming adjustments)
2. integration examples (esm/cjs).

User of the library would seek each example in different moment. It could (and often would) even be  a different developer / coder. When I'm in the middle of setting the configuration, I'd like to have different examples to be side-by-side, and don't want to be confused with integration examples. 

Because of this, and also from systematic point of view, I don't think we should have each specific example from type 1. (configuration examples) on the same level as those of the type 2. (integration examples).

To make it usable and simple for the developers, I suggest two (three) options:

#### A) Keep configuration examples in the `examples/` root, and just add integration examples
```
examples/
├── index.html
├── functions.js
├── personalization.js
├── configuration.html
├── callbacks.html
├── ...
├── webpack-with-cjs
│   ├── index.html
│   ├── index.js
│   ├── package.json
│   ├── README.md
│   └── webpack.config.js
├── webpack-with-esm
    ├── index.html
    ├── index.js
    ├── package.json
    ├── README.md
    └── webpack.config.js
```

Main pros of this is: simple access to the most used examples (configuration), so basically discoverability is the key here .
The con is obviously mixing example cases 1. with 2. in the same structure (although on different level).
However they are still separated - configuration examples right in the directory, integration examples in sub-directories. So they don't confuse with each other.

#### B) Separate configuration examples to `configuration/` directory

```
examples/
├── configuration
│   ├── callbacks.html
│   ├── configuration.html
│   ├── functions.js
│   ├── index.html
│   ├── personalization.js
│   └── theming.html
├── webpack-with-cjs
│   ├── index.html
│   ├── index.js
│   ├── package.json
│   ├── README.md
│   └── webpack.config.js
└── webpack-with-esm
    ├── index.html
    ├── index.js
    ├── package.json
    ├── README.md
    └── webpack.config.js
```

This is more systematic. However the con is that the configuration examples are more hidden in the directory structure and hard to be found.

#### C) Encapsulate integration with own directory

It would mean this structure:
```
examples/
├── configuration
└── integration
    ├── webpack-with-cjs
    └── webpack-with-esm
```

Although maybe the most systematic, I think this is a bit over-complicated, so this variant is here just for for completeness' sake :)

---

What do you think? I would be happy with variants A) and B), with A) a bit preferred because of the discoverability point.

examples/webpack-with-esm/README.md
literat: Well, Grammarly agrees but there is also missing `the` before `build` :-)

examples/with-callbacks/README.md
literat: This file was removed.

examples/with-theme/README.md
literat: This file was removed.

examples/with-extended-configuration/README.md
literat: This file was removed.

literat:
I've just taken the **A** path to refactor these examples.

examples/webpack-with-cjs/README.md
literat: Fixed.

index.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
   <meta charset="UTF-8">
-  <meta http-equiv="refresh" content="0; url=./examples/">
+  <meta http-equiv="refresh" content="0; url=./examples/browser/">
```
OndraM, line 5:
```suggestion
  <meta http-equiv="refresh" content="0; url=./examples/">
```

OndraM (changes requested):
Maybe add link to those two examples to readme.md? 

Like in the `Import the module in your javascript:` link to the esm example and in `Legacy import` link to the cjs example.

Otherwise LGTM!

literat:
> Maybe add link to those two examples to readme.md?

Please, do a quick check. I've just added them to readme.

OndraM (approved):

README.md
```diff
@@ -458,3 +462,5 @@ for more information.
 [examples-callbacks]: https://lmc-eu.github.io/cookie-consent-manager/examples/callbacks.html
 [examples-configuration]: https://lmc-eu.github.io/cookie-consent-manager/examples/configuration.html
 [examples-theming]: https://lmc-eu.github.io/cookie-consent-manager/examples/theming.html
+[examples-integration-esm]: https://github.com/lmc-eu/cookie-consent-manager/examples/webpack-with-esm/
+[examples-integration-cjs]: https://github.com/lmc-eu/cookie-consent-manager/examples/webpack-with-cjs/
```
OndraM, line 466:
```suggestion
[examples-integration-esm]: https://github.com/lmc-eu/cookie-consent-manager/tree/main/examples/webpack-with-esm/
[examples-integration-cjs]: https://github.com/lmc-eu/cookie-consent-manager/tree/main/examples/webpack-with-cjs/
```

OndraM (changes requested):

PR #59, opened 2021-11-12
Feat: Text parametrization #CCM-25

literat:
I've prepare parametrisation of translations as simple as it could be but with thoughts to future extension.

So I turned configs into function with one param - option object of messages - so we can pass down multiple messages at once.

I've also rejected parametrisation using some type of template string and replacing them by some clever function. Config is deeply nested object so I've avoided to parsing it.

In addition and to be sure that config is generated as it should be I have introduce few test using Jest Testing framework.
Configs are tested against snapshots of them. It is the simplest test and again I do not need to parse entire config object or store it and match them agains each other. But this could ve discussed.

Lastly, @OndraM should we rename `uk` config to `ua`? I am confused because `uk` is normally United Kingdom and Ukraine is `ua`.


OndraM:
> Lastly, @OndraM should we rename uk config to ua? I am confused because uk is normally United Kingdom and Ukraine is ua.

I was confused as well, but ISO 3166 country code (which is "ua" for the country Ukraine) is different than [ISO 639 language code](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) for the language Ukrainian ("uk"). 🤷

literat:
> I was confused as well, but ISO 3166 country code (which is "ua" for the country Ukraine) is different than [ISO 639 language code](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) for the language Ukrainian ("uk"). 🤷

@OndraM Well, we should definitely extend our readme, because I've just notice that newly added languages are missing. And maybe add this information to separate section. Something like "Supported languages".

OndraM:
> > I was confused as well, but ISO 3166 country code (which is "ua" for the country Ukraine) is different than [ISO 639 language code](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) for the language Ukrainian ("uk"). shrug
> 
> @OndraM Well, we should definitely extend our readme, because I've just notice that newly added languages are missing. And maybe add this information to separate section. Something like "Supported languages".

True, I will update the readme.

OndraM:
(wrong button, sorry :facepalm: )

literat:
> (wrong button, sorry 🤦 )

:D I am happy that I am not the only one to whom this happens :D

src/languages/__tests__/cs.test.js
```diff
@@ -0,0 +1,7 @@
+import { config } from '../cs';
```
OndraM, line 1: Could this be somehow more DRY to not have exactly the same test eight times?

Its been a long time I worked with Jest, but in eg. PHPUnit I would for example create a common ancestor with the common logics of the test and make eight test classes which extends it. 

Or maybe even just one test class, which will test the functions in all the eight files.

src/utils.js
```diff
@@ -0,0 +1,14 @@
...
+    }
+
+    if (i === strings.length - 1) {
+      return `${accumulator} ${and} ${string}`;
+    }
+
+    return `${accumulator}, ${string}`;
+  });
+
+export const plural = (count, singular, plural) => (count === 1 ? singular : plural);
```
OndraM, line 14: What about call it `pluralize`? This would sound as a more descriptive name to me.

src/utils.js
```diff
@@ -0,0 +1,14 @@
+export const addSeparators = (strings, and) =>
+  strings.reduce((accumulator, string, i) => {
+    if (i === 0) {
+      return `${accumulator}${string}`;
+    }
+
+    if (i === strings.length - 1) {
+      return `${accumulator} ${and} ${string}`;
```
OndraM, line 8:
```suggestion
      return `${accumulator} ${and} ${string}`;
```
(replace ordinary space with no-break space character ` `)

src/languages/__tests__/cs.test.js
```diff
@@ -0,0 +1,7 @@
...
+
+describe('config', () => {
+  it('should return localization object', () => {
+    expect(config({ companyNames: ['test1', 'test2', 'test3'] })).toMatchSnapshot();
+  });
+});
```
OndraM, line 7: Proof of concept for table testing, so that all configs are checked using one test to DRY:

```suggestion
import { config as configCs } from '../cs';
import { config as configDe } from '../de';

describe.each([
  ['cs', configCs],
  ['de', configDe]
])('%s config', (language, config) => {
  it('should return localization object', () => {
    expect(config({companyNames: ['test1', 'test2', 'test3']})).toMatchSnapshot();
  });
})
```

src/utils.js
literat: Fixed.

src/utils.js
literat: Fixed.

src/languages/__tests__/cs.test.js
literat: Yeah, I know that this is something I should refactor. But I did not know if this concept of testing would be appropriate and if you would ok with it. 

I have refactored it using `jest-each`. But still, if you update translations in the future it would be also needed to update snapshots as well using `yarn test:unit -u`.

src/languages/__tests__/cs.test.js
literat: Refactored.

OndraM (changes requested):

src/languages/__tests__/langConfig.test.js
```diff
@@ -0,0 +1,24 @@
...
+import { config as configCs } from '../cs';
+import { config as configDe } from '../de';
+import { config as configEn } from '../en';
+import { config as configHu } from '../hu';
+import { config as configPl } from '../pl';
+import { config as configRu } from '../ru';
+import { config as configSk } from '../sk';
+import { config as configUk } from '../uk';
+
+each([
```
OndraM, line 11: BTW why you used `each([]).describe()` instead of `describe.each([])()` from [jest](https://jestjs.io/docs/api#1-describeeachtablename-fn-timeout)?

literat: Good point, this is something new to me because when I started to use `jest-each` there was no native `.each` helper. Thanks. Refactored.

src/languages/cs.js
```diff
@@ -1,56 +1,80 @@
...
+  return {
+    consent_modal: {
+      title: 'Tyto stránky využívají cookies',
+      description: `Kliknutím na „Přijmout vše“ dáváte souhlas ${pluralize(
+        lang.companyNames.length,
+        lang.company,
+        lang.companies,
+      )} ${addSeparators(
+        lang.companyNames,
+        extra.and,
```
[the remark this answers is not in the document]
literat: Fixed.

src/languages/cs.js
```diff
@@ -1,56 +1,80 @@
...
+
+const extra = {
+  and: 'a',
+  company: 'společnosti',
+  companies: 'společnostem',
+};
+
+/**
+ * @param {Object} [messages] - Object with extra messages
+ * @param {array} [args.companyNames] - Array of strings with company names used to parametrized translations
```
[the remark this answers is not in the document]
literat: Fixed.

src/LmcCookieConsentManager.js
```diff
@@ -31,6 +32,7 @@ const defaultOptions = {
  * @param {function} [args.onAccept] - Callback to be executed when any consent is detected (either given right now or already saved previously)
  * @param {function} [args.onAcceptOnlyNecessary] - Callback to be executed when consent with only necessary cookies is detected (either given right now or already saved previously)
  * @param {function} [args.onAcceptAll] - Callback to be executed when consent with all cookies is detected (either given right now or already saved previously)
+ * @param {array} [args.companyNames] - Array of strings with company names used to parametrized translations
```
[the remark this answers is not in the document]
literat: Fixed

README.md
```diff
@@ -206,6 +206,7 @@ initLmcCookieConsentManager( // when loaded as a module, these options are passe
 |---------------|----------|--------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------|
 | `defaultLang` | string   | 'cs'                           | Default language. One of `cs`, `de`, `en`, `hu`, `pl`, `ru`, `sk`, `uk`. This language will be used when autodetect is disabled or when it fails.                 |
 | `autodetectLang`| string | true                           | Autodetect language from the browser. If autodetect fails or if unsupported language is detected, fallback to `defaultLang`.<br>When disabled, force language to `defaultLang`. |
+| `companyNames`| array | ['LMC']                           | Array of strings with company names used to parametrized translations. |
```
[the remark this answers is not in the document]
literat: Fixed

OndraM:
Looks nice! So could you un-draft the PR (and rebase to `main`)?

OndraM (approved):
:shipit:

PR #61, opened 2021-11-14
Docs: Do not use patch part in URL to update to bugfix versions automatically

OndraM:
Ref. [#CCM-29](https://jira.int.lmc.cz/browse/CCM-29)

Using this users who will use the "recommended" URL to CDN will get bugfixes.

I also think after 1.0.0 we should also drop the "minor" version from the URL. We comply with semver, so this will mean they will get the latest compatible version. But this is up to discussion, as this might possibly be a bit more risky.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,13 +13,13 @@ Load default CSS in your `<head>`:
 
 ```html
 <link rel="preconnect" href="https://cdn.jsdelivr.net">
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@lmc-eu/cookie-consent-manager@0.5.0/LmcCookieConsentManager.min.css">
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@lmc-eu/cookie-consent-manager@0.5/LmcCookieConsentManager.min.css">
 ```
 
 Load the script and initialize the plugin right before ending `</body>` tag:
 
 ```html
-<script defer src="https://cdn.jsdelivr.net/npm/@lmc-eu/cookie-consent-manager@0.5.0/init.js"></script>
+<script defer src="https://cdn.jsdelivr.net/npm/@lmc-eu/cookie-consent-manager@0.5/init.js"></script>
 <script>
 window.addEventListener('load', function () {
   initLmcCookieConsentManager();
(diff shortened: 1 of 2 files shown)


scripts/readme-replace-version.js
```diff
@@ -1,9 +1,12 @@
 const replace = require('replace-in-file');
 const packageJson = require('../package.json');
+// TODO: replace with '$1' after 1.0 release
+const newVersion = packageJson.version.replace(/([0-9])\.([0-9])\.([0-9])/, '$1.$2');
```
literat, line 4: Well, can also just `match` the prefered part and get first match as `.match(/([0-9])\.([0-9]/)[0]`. No need for match and replace.

And after this little refactor you can also add the regex into variable and use it on both places here, i think ;-)

README.md
```diff
@@ -53,8 +53,8 @@ as shown above or to explicitly specify the desired font yourself (head to [Them
 You can load the plugin from a CDN, as in the basic example above.
 
 ```html
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@lmc-eu/cookie-consent-manager@0.5.0/LmcCookieConsentManager.min.css">
-<script defer src="https://cdn.jsdelivr.net/npm/@lmc-eu/cookie-consent-manager@0.5.0/init.js"></script>
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@lmc-eu/cookie-consent-manager@0.5/LmcCookieConsentManager.min.css">
+<script defer src="https://cdn.jsdelivr.net/npm/@lmc-eu/cookie-consent-manager@0.5/init.js"></script>
```
literat, line 57: Please, add a simple note/disclaimer here why we are using "latest" version. It should be clear to everyone but, you know, sometimes is everbody suprised with some version with unintenional bug in it.

literat (changes requested):

scripts/readme-replace-version.js
OndraM: Changed to match, but I'm not quite sure it is more readable :upside_down_face: : .

Also I don't think it will not be good to reuse the regexp below, as we are looking for different pattern and concatenating the regexp would make it less clear, IMHO.

README.md
OndraM: Added, have a look.

literat (approved):

PR #68, opened 2021-11-16
Require service name to be passed as the first param

OndraM:
This is the first step of CCM-35. We will require service name to be given on initialization, so that we could submit the consents (CCM-13) in near future, without breaking the API later.

### DIFF
--- a/src/LmcCookieConsentManager.js
+++ b/src/LmcCookieConsentManager.js
@@ -22,6 +22,7 @@ const defaultOptions = {
 };
 
 /**
+ * @param {string} serviceName - Identifier of the source service (website/application). Must be provided.
  * @param {Object} [args] - Options for cookie consent manager
  * @param {string} [args.defaultLang] - Default language. Must be one of predefined languages.
  * @param {boolean} [args.autodetectLang] - Autodetect language from the browser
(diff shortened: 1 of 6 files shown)


src/LmcCookieConsentManager.js
```diff
@@ -34,7 +35,11 @@ const defaultOptions = {
  * @param {Object} [args.config] - Override default config. See https://github.com/orestbida/cookieconsent/blob/master/Readme.md#all-available-options
  * @returns {Object} Instance of the underlying CookieConsent component. For available API, see https://github.com/orestbida/cookieconsent#apis--configuration-parameters
```
literat, line 36: Please, document that it throws something using`@throws`. :-)

src/LmcCookieConsentManager.js
```diff
@@ -34,7 +35,11 @@ const defaultOptions = {
...
  */
-const LmcCookieConsentManager = (args) => {
+const LmcCookieConsentManager = (serviceName, args) => {
+  if (!serviceName || serviceName === '' || typeof serviceName !== 'string') {
+    throw new Error('serviceName is a required parameter');
```
literat, line 40: Well, I think that the message is misleading when I provide empty string or number :-) Maybe `serviceName is a required parameter and should be type of string!`

literat (changes requested):

src/LmcCookieConsentManager.js
OndraM: Added, however, I'm really not a fan of documenting behavior, which users of the method should not care about and don't need to handle. This IMHO only adds useless noise and clutter and has no value for the users. Because what information it gives you? Information about the required parameter and its type is already in `@param`, so this is just redundant.  I guess we don't want anyone to encapsulate this into try-catch, so that this exception ("Parameter not provided") could be handled...

literat: Ok, you are right.

literat (approved):

src/LmcCookieConsentManager.js
literat: Let it there or remove it, it is up to you.

literat (approved):

PR #70, opened 2021-11-17
Feat: Generate and store consent UUID on accept

OndraM:
The next part of CCM-35 after #69.

I thought we would not need uuid library... However from what I've read regarding Math.random(), any uuid function based on it would mean high chance of collisions. And since the traffic this would have and necessity to have uuid really unique, this could be problematic.

### DIFF
--- a/src/LmcCookieConsentManager.js
+++ b/src/LmcCookieConsentManager.js
@@ -1,4 +1,5 @@
 import 'vanilla-cookieconsent';
+import { nanoid } from 'nanoid';
 
 import { config as configCs } from './languages/cs';
 import { config as configDe } from './languages/de';
@@ -91,6 +92,14 @@ const LmcCookieConsentManager = (serviceName, args) => {
       onAccept(cookie, cookieConsent);
 
       if (isFirstTimeAccept) {
+        const cookieData = cookieConsent.get('data');
+        if (cookieData === null || !('uid' in cookieData)) {
+          cookieConsent.set('data', {
+            value: { serviceName: serviceName, uid: nanoid() },
+            mode: 'update',
+          });
+        }
+
         onFirstAccept(cookie, cookieConsent);
         acceptedOnlyNecessary
           ? onFirstAcceptOnlyNecessary(cookie, cookieConsent)
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
   },
   "dependencies": {
     "@lmc-eu/spirit-design-tokens": "^0.4.0",
+    "nanoid": "^3.1.30",
     "vanilla-cookieconsent": "^2.6.1"
   },
   "devDependencies": {


src/LmcCookieConsentManager.js
```diff
@@ -91,6 +92,13 @@ const LmcCookieConsentManager = (serviceName, args) => {
       onAccept(cookie, cookieConsent);
 
       if (isFirstTimeAccept) {
+        if (cookieConsent.get('data') === null || !('uuid' in cookieConsent.get('data'))) {
```
OndraM, line 95: The second condition is for possibility someone will store something in the cookie as part of onAccept above, and we still need to generate and store the UUID.

src/LmcCookieConsentManager.js
```diff
@@ -1,4 +1,5 @@
 import 'vanilla-cookieconsent';
+import { v4 as uuidv4 } from 'uuid';
```
literat, line 2: Please, consider use of `nanoid` instead of `uuid`. It is faster, smaller and uses better random number generator: https://github.com/ai/nanoid#comparison-with-uuid

literat (changes requested):

src/LmcCookieConsentManager.js
OndraM: Nice, didn't know this library - replaced.

literat: I think we have a conversation about this already - https://lmccz.slack.com/archives/CU9D52MTK/p1625815872001000 ;-)

src/LmcCookieConsentManager.js
literat: Maybe little optimization but should't it `cookieConsent.get('data')` be in a variable? Because in that condition it is used twice.

literat (changes requested):

src/LmcCookieConsentManager.js
OndraM: Ok, I thought it internally just returns the value, but it actually parses the json once again, so ok :upside_down_face:  , although it is really just a microoptimalization :)

src/LmcCookieConsentManager.js
```diff
@@ -91,6 +92,14 @@ const LmcCookieConsentManager = (serviceName, args) => {
       onAccept(cookie, cookieConsent);
 
       if (isFirstTimeAccept) {
+        const cookieData = cookieConsent.get('data');
+        if (cookieData === null || !('uuid' in cookieData)) {
+          cookieConsent.set('data', {
+            value: { serviceName: serviceName, uuid: nanoid() },
```
literat, line 98: In a case when `cookieData` is not null but there is no `uuid` property you are replacing entire value object. Maybe it will be safer if you destructure old value into new one.

`value: { ...cookieData.value, serviceName: serviceName, uuid: nanoid() },`

But there is a problem with existence of property `value` on a `null`. 🤔

literat (changes requested):

src/LmcCookieConsentManager.js
OndraM: > In a case when cookieData is not null but there is no uuid property you are replacing entire value object. 

Not really, see `mode: 'update'`. This [is handled](https://github.com/orestbida/cookieconsent/blob/master/src/cookieconsent.js#L1353-L1370) by the vanilla library and it should only update the given properties.

literat: Thanks, I did not notice.

literat (approved):

OndraM:
One more change - replace `uuid` which refers to the "Universally unique identifier" standard (we no longer use) with `uid` as more generic "user id"...

literat (approved):

PR #71, opened 2021-11-18
Chore/ccm 6 codestyle linting

literat:
Introducing code style linting using ESLint and our brand new @lmc-eu/eslint-config-base rules. Just to try how it will behave.


examples/personalization.js
```diff
@@ -1,3 +1,4 @@
 // This mocks external script which provides personalized content (and uses cookies to identify the user).
 // As such, it should only by run when user has given consent for "personalization" category.
-document.getElementById('personalization-content').innerText = 'personalization.js loaded and executed';
+document.getElementById('personalization-content').innerText =
```
OndraM, line 3: What rule is applied here? It has 105 characters and [there is 120 characters limit](https://github.com/lmc-eu/code-quality-tools/pull/26/files#diff-870a1e4feddd30425c04bb8a2a0b1638b5d28e418acda7f8ab2dcd7727b92bbbR37-R43) :thinking:

src/LmcCookieConsentManager.js
```diff
@@ -26,11 +27,15 @@ const defaultOptions = {
...
+ *                                                       are accepted
  * @param {function} [args.onFirstAcceptAll] - Callback to be executed right after all cookies are accepted
- * @param {function} [args.onAccept] - Callback to be executed when any consent is detected (either given right now or already saved previously)
- * @param {function} [args.onAcceptOnlyNecessary] - Callback to be executed when consent with only necessary cookies is detected (either given right now or already saved previously)
- * @param {function} [args.onAcceptAll] - Callback to be executed when consent with all cookies is detected (either given right now or already saved previously)
+ * @param {function} [args.onAccept] - Callback to be executed when any consent is detected (either given right now
+ *                                     or already saved previously)
+ * @param {function} [args.onAcceptOnlyNecessary] - Callback to be executed when consent with only necessary cookies
+ *                                                  is detected (either given right now or already saved previously)
```
OndraM, line 36: Just a personal preference, but this seems ugly to me :smile:  .

Cannot it break it like at least like this?

```suggestion
 * @param {function} [args.onAcceptOnlyNecessary] - Callback to be executed when consent with only necessary cookies
 * is detected (either given right now or already saved previously)
```

Or maybe ignore jsdoc altogether? :shrug:

src/init.js
```diff
@@ -1,6 +1,6 @@
 import LmcCookieConsentManager from './LmcCookieConsentManager';
 
-(function () {
+(function initialize() {
```
OndraM, line 3: Why is this style better, does it have any implications? 🤔

src/LmcCookieConsentManager.js
literat: Yes, this is caused by `max-len` and I have a comment there: `ignoreComments: false, // @TODO: should we also ignore comments?`. I have no preference here but line is way too long (I must turn my editor into distraction free mode to see it whole).

src/init.js
literat: In a short, it is better for debugging purposes. You do not want to see `anonymous function` in stack trace.

For more see https://eslint.org/docs/rules/func-names

src/LmcCookieConsentManager.js
OndraM: I don't have strong preference either, but this line-break mode (with space aligning from the left) just seems unpractical. But :shrug:   

However wouldn't one read the function signature mainly in the IDE? It handles the jsdoc parsing, line breaks, formatting etc. for you...

obrazek

examples/personalization.js
literat: This is caused by [eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier) which runs prettier as an eslint rule and we have set `printWidth` [recently](https://github.com/lmc-eu/code-quality-tools/commit/f074cacccd4b670af800ecaefe247105575190fc) to 100. I think it is your commit :-)

src/LmcCookieConsentManager.js
OndraM: Related - cannot we somehow inherit jsdoc from this `LmcCookieConsentManager` to `initLmcCookieConsentManager`? Because if someone uses it via `initLmcCookieConsentManager()`, there won't be any jsdoc and type hinting from the IDE at all :thinking: 

obrazek

examples/personalization.js
OndraM: Yes, but `printWidth` is soft-limit, it should include some heuristic to not do this kind of breaks... At least I thought :-D .

BTW, doesn't this mean it runs prettier twice? (Not on this file obviously, but on the files in src/). Once as part of `format:check`, and then in `js:lint` as part of eslint.

literat: Well, we can override this to be same as 
```javascript
module.exports = {
  ...require('@lmc-eu/prettier-config'),
  printWidth: 120,
};
```
Because this is somehow conflicting rule and eslint-plugin-prettier makes it to be hard limit.

Yes, in this case prettier runs twice. So we should disable `format:check` or eslint-plugin-prettier.

src/LmcCookieConsentManager.js
literat: Hmm, have a look here: https://stackoverflow.com/questions/30676286/what-is-the-proper-canonical-formatting-of-long-jsdoc-lines

I will prefer the Google way. Not everyone have perfectly configured IDE and I think that the code should be readable mainly in the raw form.

OndraM: Agreed, the "Google" way with "one tab" indent looks fine, but thats not what its here. This could lead to a lot of short lines, as mentioned in the SO question.

literat: Now I am quite confused. Should we use **google way**
```
 * @param {function} [args.onFirstAcceptOnlyNecessary] - Callback to be executed right after only necessary cookies
 *   are accepted
```
or
**javadoc way**
```
 * @param {function} [args.onFirstAcceptOnlyNecessary] - Callback to be executed right after only necessary cookies
                                                         are accepted
```
or something different?

OndraM: "google way" - so now it looks fine, sorry, I missed you already updated the PR :)

examples/personalization.js
OndraM: Ok, well, then i think we should unify the limit everywhere to 120 characters, as we discussed yesterday... Will you please update the prettier config (maybe include it in https://github.com/lmc-eu/code-quality-tools/pull/26 )? Or should I send a separate PR?

package.json
```diff
@@ -37,9 +37,10 @@
     "css:lint": "stylelint --config .stylelintrc \"src/**/*.scss\" --cache --cache-location .cache/.stylelintcache",
     "prejs": "yarn js:lint",
     "js": "yarn js:compile",
-    "js:lint": "echo 'To be defined'",
+    "js:lint": "eslint ./src ./examples ./scripts",
```
adamkudrna, line 40: Root config files should be included too:

```suggestion
    "js:lint": "eslint ./*.js ./src ./examples ./scripts",
```

src/LmcCookieConsentManager.js
```diff
@@ -92,9 +97,7 @@ const LmcCookieConsentManager = (args) => {
           : onFirstAcceptAll(cookie, cookieConsent);
       }
 
-      acceptedOnlyNecessary
-        ? onAcceptOnlyNecessary(cookie, cookieConsent)
-        : onAcceptAll(cookie, cookieConsent);
+      acceptedOnlyNecessary ? onAcceptOnlyNecessary(cookie, cookieConsent) : onAcceptAll(cookie, cookieConsent);
```
adamkudrna, line 100: Why doesn't ESlint like this? It was better readable before, IMO.

adamkudrna (approved):

OndraM:
@literat FYI apart from the few comments above this seems ok to me, so just ping me or re-request review once this ready to final review :)

examples/personalization.js
literat: Print width increased to 120: https://github.com/lmc-eu/code-quality-tools/pull/29

package.json
literat: Fixed.

src/LmcCookieConsentManager.js
literat: This was some bas formatting using eslint + prettier. Now it is ok.

src/consentCollector.js
```diff
@@ -58,4 +45,17 @@ const postDataToApi = async (apiUrl, payload) => {
   return response.json();
 };
 
+/**
```
OndraM, line 48: What rule applies here? 🤔

literat: This was solved manually: https://eslint.org/docs/rules/no-use-before-define

The second one was `Using exported name 'submitConsent' as identifier for default export  import/no-named-as-default`: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-named-as-default.md

OndraM: > https://eslint.org/docs/rules/no-use-before-define

Is this rule really needed for us, or is there because of just some legacy reasons? Because this method order seems quite unintuitive to me - start the file with something I consider implementation details - some low-level "internal" methods (buildPayload, postDataToApi...) while having the "high-level public API" (what is the `submitConsent` method) of this import defined somewhere as the last method of the file. This rules makes it impossible to order the methods by such logic...

I read the documentation, but I did not understand in what scenario could the current functions order cause any actual trouble...? Or is is just a convention?

literat: Exact definition of our rule is here: https://github.com/airbnb/javascript/blob/1eadb93e377da1e56c3f91f26610e5d0a00738a9/packages/eslint-config-airbnb-base/rules/variables.js#L54. We are extending what I consider as a broad community standard which AirBnB is.

Our problem is  difference between  function declaration and function expression (which in our case is arrow function expression). So in fact this rule is applied because we are using "variables". And to have variable defined before I use it make great sense to me. And also because function expressions are not hoisted (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/function).

To have structure of function as you suggested I recommend to refactor those function into function declaration, eg. `const submitConsent = (...)` -> `function submitConsent (...)`.

OndraM: I leave up to you to choose the best practice. So feel free to merge/refactor this, based o what you think is the best. Mine feedback is just that this functions order looks weird and unintutitive to me, from the high- vs. low- level function. But 🤷

literat: Because that function declaration is safe, I have configured that rule not to apply on function declaration. And also refactored those function to function declaration and top up the main function.

PR #73, opened 2021-11-19
Feat: Submit consent to the API

OndraM:
See [#CCM-39](https://jira.int.lmc.cz/browse/CCM-39) .

Will be opt-in first. Once tested, the `enableConsentSubmit` configuration will either by opt-out or completely removed.

Naming suggestions welcomed! (I'm not sure about `dataCollector`, `enableConsentSubmit` and `submitConsent` names...)

### DIFF
--- a/examples/configuration.html
+++ b/examples/configuration.html
@@ -80,6 +80,7 @@
         autodetectLang: false, // do not detect language from the browser
         defaultLang: 'en', // use 'en' language by default
         companyNames: ['LMC', 'Some Other Company', 'ACME'], // define custom company names to be shown in the consent text
+        consentCollectorApiUrl: 'https://ccm.lmc.cz/local-data-acceptation-data-entries?Spot=(public,demo)', // override default URL for demo purposes; do not set custom consentCollectorApiUrl unless you have been explicitly guided to do so!
         config: { // override default config of the underlying library, see https://github.com/orestbida/cookieconsent#all-available-options
           delay: 500, // show cookie consent banner after 500ms
           page_scripts: false, // disable third-party script management, see https://github.com/lmc-eu/cookie-consent-manager#third-party-scripts-loaded-via-script
--- a/src/LmcCookieConsentManager.js
+++ b/src/LmcCookieConsentManager.js
@@ -9,10 +9,12 @@ import { config as configPl } from './languages/pl';
 import { config as configRu } from './languages/ru';
 import { config as configSk } from './languages/sk';
 import { config as configUk } from './languages/uk';
+import submitConsent from './consentCollector';
 
 const defaultOptions = {
   defaultLang: 'cs',
   autodetectLang: true,
+  consentCollectorApiUrl: 'https://ccm.lmc.cz/local-data-acceptation-data-entries',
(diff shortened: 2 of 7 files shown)


examples/configuration.html
```diff
@@ -79,6 +79,7 @@ <h3 class="mt-4">Use cookieConsent instance</h3>
       {
         autodetectLang: false, // do not detect language from the browser
         defaultLang: 'en', // use 'en' language by default
+        enableConsentSubmit: true,
```
literat, line 82: Missing comment?

src/LmcCookieConsentManager.js
```diff
@@ -9,10 +9,12 @@ import { config as configPl } from './languages/pl';
 import { config as configRu } from './languages/ru';
 import { config as configSk } from './languages/sk';
 import { config as configUk } from './languages/uk';
+import submitConsent from './dataCollector/dataCollector';
```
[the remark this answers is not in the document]
literat: Are expecting that `dataCollector` library will include more files? Maybe it will be better to have `dataCollector.js` in root an export only one function like `import {submitConsent} from './dataCollector'`.

src/dataCollector/dataCollector.js
```diff
@@ -0,0 +1,62 @@
...
+      },
+    },
+  };
+};
+
+/**
+ * @param {Object} payload
+ * @return {Promise<any>}
+ */
+const postDataToApi = async (payload) => {
```
literat, line 46: Maybe `sendCookieConsentToCollector`? `data` is something general and api also.

src/LmcCookieConsentManager.js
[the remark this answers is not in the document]
OndraM: > Are expecting that `dataCollector` library will include more files?

I think not. 

But I'm used not to pollute the root directory with other files except some "main". But the convention here could be different, so I don't mind 🤷. So why not, I will move it.

src/dataCollector/dataCollector.js
```diff
@@ -0,0 +1,62 @@
...
+  const dataCollectorUrl =
```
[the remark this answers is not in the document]
MortalFlesh: technically.. it is `aclDataEntryCCMUrl` but name it as you please :)

src/dataCollector.js
```diff
@@ -0,0 +1,62 @@
...
+  };
+};
+
+/**
+ * @param {Object} payload
+ * @return {Promise<any>}
+ */
+const postDataToApi = async (payload) => {
+  const dataCollectorUrl =
+    'https://ccm.lmc.cz/local-data-acceptation-data-entries?Spot=(arch,test)'; // TODO: remove test spot after beta
```
MortalFlesh, line 48: It could be easier to use (and test) if the spot would be a public option - so a client could set it.

src/dataCollector/dataCollector.js
OndraM: I understand, but this method does not have any logic related to consent. It just receives data (whatever they are) and POST them...

literat (approved):

MortalFlesh (approved):

literat (approved):

PR #78, opened 2021-11-22
Refactor: Extract assemble config into variable

OndraM:
The length of `run()` call is now >50 lines long, so this is hard to read and code smell.

Also, I sometimes want to dump the assembled config, which is now much easier.

(BTW In the future I think it may also be inevitable to move the whole config object assembly to custom method.)

### DIFF
--- a/src/LmcCookieConsentManager.js
+++ b/src/LmcCookieConsentManager.js
@@ -71,7 +71,7 @@ const LmcCookieConsentManager = (serviceName, args) => {
     uk: configUk({ companyNames }),
   };
 
-  cookieConsent.run({
+  const cookieConsentConfig = {
     auto_language: autodetectLang, // Enable detection from navigator.language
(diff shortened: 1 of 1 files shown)


src/LmcCookieConsentManager.js
```diff
@@ -71,7 +71,7 @@ const LmcCookieConsentManager = (serviceName, args) => {
     uk: configUk({ companyNames }),
   };
 
-  cookieConsent.run({
+  const cookieConsentConfig = {
```
literat, line 74: If you are refactoring this do also `const handleAccept = (cookie) => {...}` for our `onAccept` function, please. In config then stays only `onAccept: handleAccept, `. :-)

literat (changes requested):

src/LmcCookieConsentManager.js
literat: Ok.

literat (approved):

PR #80, opened 2021-11-23
Feat: Option to define URL to the API

OndraM:
This is a variant of #73. This adds option to use custom API url, and also an option to disable it by providing null. 

The use cases are:
* default - send consents, will use `https://ccm.lmc.cz/local-data-acceptation-data-entries`
* disable sending - use `null`
* use custom URL / Spot - provide `https://ccm.lmc.cz/local-data-acceptation-data-entries?Spot=(arch,test)`

Sending separately, as this is kind of "request for comments".


OndraM:
Ok, after further discussions the seems ok. So closing this and adding into #73.

PR #93, opened 2021-11-26
Docs/improve font instructions

adamkudrna:
- Fix: Revert font family inheritance as it blocked font customization
- Feat: Let base font size be inherited from surrounding text
- Docs: Make clear that Inter font link is mandatory unless font family is customized
- Deps: Update Spirit Design Tokens

As a result:

- Font size should work normally even on sites that use 10px hack for easier calculations.
- Cookie consent UI is rendered with a sans-serif font until Inter web font is loaded.
- We make it clear that Inter font is mandatory unless font family is customized (either via Spirit or custom properties).

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,6 +317,8 @@ anywhere in your stylesheet (the order of which is not important):
 }
 ```
 
+[👀 See theming example][examples-theming]
+
 <details>
 <summary>Full list of available CSS custom properties:</summary>
 
@@ -439,3 +441,4 @@ for more information.
 [examples]: https://lmc-eu.github.io/cookie-consent-manager/examples/
 [examples-callbacks]: https://lmc-eu.github.io/cookie-consent-manager/examples/callbacks.html
 [examples-configuration]: https://lmc-eu.github.io/cookie-consent-manager/examples/configuration.html
+[examples-theming]: https://lmc-eu.github.io/cookie-consent-manager/examples/theming.html
--- a/README.md
+++ b/README.md
@@ -9,10 +9,18 @@ The package is a wrapper around [Cookie Consent] by [Orest Bida].
 
 ## Basic usage
 
-Load default CSS in your `<head>`:
+Make assets load faster by placing pre-connect headers right after `<meta charset>` in your `<head>`:
 
 ```html
 <link rel="preconnect" href="https://cdn.jsdelivr.net">
+<link rel="preconnect" href="https://fonts.googleapis.com">
+<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+```
+
+Load default CSS along with your styles in `<head>`:
+
+```html
+<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Inter:wght@400;700&display=swap">
 <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@lmc-eu/cookie-consent-manager@0.7/LmcCookieConsentManager.min.css">
 ```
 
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -259,7 +254,10 @@ All you need to do is to add this plugin's SCSS to your Sass pipeline:
 @use '@lmc-eu/cookie-consent-manager/LmcCookieConsentManager';
 ```
 
-Now set up a [Sass load path] so the Sass compiler can find stylesheets located in the `node_modules` directory
+<details>
+<summary>Make sure you have `node_modules` and path to your design tokens in your Sass include paths.</summary>
```
OndraM, line 258:
```suggestion
<summary>Make sure you have <code>node_modules</code> and path to your design tokens in your Sass include paths.</summary>
```

The markdown obviously does not work here 🤷

README.md
```diff
@@ -289,15 +287,24 @@ Or with webpack:
...
 
+</details>
+
 **Note:** `sass` v1.23 or higher is required to be able to compile the new Sass modules syntax. You may need to migrate
 to [`sass`][sass] since all other Sass compilers (and the old `@import` rule) are now [deprecated][sass modules].
 
 ### Without Spirit Design System
 
+Handful of CSS custom properties are available for you to customize the UI.
+Change their values to match the design of your site, for example:
```
OndraM, line 298: Maybe describe some theory of operation - how does it work in this scenario. That you must load the default css, and then use CSS variables to customize it.

README.md
```diff
@@ -92,17 +85,19 @@ via npm package [@lmc-eu/cookie-consent-manager](https://www.npmjs.com/package/@
...
-   @use 'node_modules/@lmc-eu/cookie-consent-manager/LmcCookieConsentManager.css';
+   @import url("https://fonts.googleapis.com/css2?family=Inter:wght@400;700&display=swap");
+   @use "node_modules/@lmc-eu/cookie-consent-manager/LmcCookieConsentManager.css";
    ```
 
    Please mind the `.css` extension used in the Sass example. Using the provided `.scss` stylesheet is
    recommended only for projects that are built [with Spirit Design System](#with-spirit-design-system).
 
-   See below for [theming customization options](#theming).
+   See below for [theme customization options](#theming).
```
OndraM, line 100: As we discussed personally, maybe extract font-related information to step 4, so that the installation is more gradual. (1. npm install; 2. import js; 3. import css; 4. import font).

OndraM (changes requested):

README.md
```diff
@@ -321,22 +342,39 @@ to [`sass`][sass] since all other Sass compilers (and the old `@import` rule) ar
...
+
+In both cases, following tags may be redundant and you should consider removing them from your `<head>`, should you
+have initially followed the [basic](#basic-usage) instructions:
+
+```html
+<!-- Remove `preconnect` headers only if you are NOT loading your custom font via Google Fonts: -->
+<link rel="preconnect" href="https://fonts.googleapis.com">
+<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+<!-- Remove Inter font link: -->
+<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Inter:wght@400;700&display=swap">
```
OndraM, line 368: I would shorten this whole block just to some sentence like:

```suggestion
If you used custom font like this, make sure you don't load the default Inter font and that you use `<link rel="preconnect" ...>` only to actually needed domains.
```

Because if you followed the new installation scenario, you would not came to the case when you first added the font and then you need to remove it. So this is imho a redundant and overly-detailed information.

Yes, loading the font is in the basic usage, but I see it as a quick information how does this work and how to get it quickly working, not as a full copy-paste code someone will put to production code, without reading the rest from the readme... 🤷

OndraM (approved):

README.md
adamkudrna: It's not clear to me even as an author of the docs. But if you think that _Basic usage_ is more about how it works and not about how people should actually use it, let's go ahead and simplify it as you suggest.

OndraM: > It's not clear to me even as an author of the docs.

:smile:  :thinking: 


Well, this was just a note. For me has this part opposite meaning, it would confuse me: "Why is it describing in such detailed way how to remove something I didn't even add?"

But if you think this would help you if you would be in the role of someone implementing this, just decide yourselves whether keep this or not. Maybe at least reverse order of the sentence, to begin with "If you initially followed the [basic](#basic-usage) instructions, ...". So that you know whether it is for you relevant or not.

adamkudrna: I changed it as you originally suggested. Let's wait and see what developers say once they start using it.

PR #94, opened 2021-11-26
Docs/custom theme demo

adamkudrna:
Snímek obrazovky 2021-11-26 v 22 18 54
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,6 +317,8 @@ anywhere in your stylesheet (the order of which is not important):
 }
 ```
 
+[👀 See theming example][examples-theming]
+
 <details>
 <summary>Full list of available CSS custom properties:</summary>
 
@@ -439,3 +441,4 @@ for more information.
 [examples]: https://lmc-eu.github.io/cookie-consent-manager/examples/
 [examples-callbacks]: https://lmc-eu.github.io/cookie-consent-manager/examples/callbacks.html
 [examples-configuration]: https://lmc-eu.github.io/cookie-consent-manager/examples/configuration.html
+[examples-theming]: https://lmc-eu.github.io/cookie-consent-manager/examples/theming.html
--- a/README.md
+++ b/README.md
@@ -322,8 +322,8 @@ anywhere in your stylesheet (the order of which is not important):
 
 | CSS custom property                   | Description                                             |
 |---------------------------------------|---------------------------------------------------------|
-| `--lmcccm-base-font-size`             | Base font size                                          |
 | `--lmcccm-font-family`                | Base font family                                        |
+| `--lmcccm-base-font-size`             | Base font size                                          |
 | `--lmcccm-bg`                         | Bar background color                                    |
 | `--lmcccm-text`                       | Text color                                              |
 | `--lmcccm-link-text`                  | Link text color                                         |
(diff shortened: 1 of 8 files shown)


examples/theming.html
```diff
@@ -2,19 +2,62 @@
...
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/prismjs@v1.25/themes/prism.css">
+    <!-- For the purpose of this demo, we are loading two typefaces: -->
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Inter:wght@400;700&family=Noto+Sans:wght@400;700&display=swap">
     <script src="./functions.js" defer></script>
 
-    <link rel="stylesheet" href="../dist/LmcCookieConsentManager.css">
+    <link rel="stylesheet" href="../dist/LmcCookieConsentManager.min.css"><!-- You will load this from CDN instead -->
+
+    <style type="text/css">
+        .ikea-theme {
```
OndraM, line 22: 🤔 Maybe do not use this name to avoid trademark issues 🤷

OndraM (changes requested):
👍 Nice to see the customization example!

literat (approved):

OndraM (approved):

examples/theming.html
```diff
@@ -52,19 +95,70 @@ <h1>LMC Cookie Consent Manager</h1>
...
+    --lmcccm-btn-primary-active-bg: rgb(0, 0, 0);
+    --lmcccm-btn-primary-active-text: var(--lmcccm-btn-primary-text);
+    --lmcccm-btn-secondary-text: var(--lmcccm-link-text);
+    --lmcccm-btn-secondary-hover-text: var(--lmcccm-link-hover-text);
+    --lmcccm-btn-secondary-active-text: var(--lmcccm-link-active-text);
+}</code></pre>
+            </details>
+            <p>
+                <button type="button" class="btn btn-outline-primary" onclick="document.documentElement.classList.toggle('custom-theme')">
+                    Toggle custom theme
```
OndraM, line 137: While the dark mode icon is state-full, this one is not. Also because of this, it was not obvious to me these two things could be combined (custom theme + dark mode). 

Maybe the best UI element for these two things would be something like checkbox or some on-off switch.

adamkudrna: I've tried switch and it seemed too small me… I'll give it another try.

adamkudrna: We agreed on moving this to a standalone PR so this one is not blocked.

README.md
```diff
@@ -303,29 +303,34 @@ Change their values to match the design of your site, for example:
...
-| `--lmcccm-base-font-size`             | Base font size                                          |
 | `--lmcccm-font-family`                | Base font family                                        |
+| `--lmcccm-base-font-size`             | Base font size                                          |
 | `--lmcccm-bg`                         | Bar background color                                    |
 | `--lmcccm-text`                       | Text color                                              |
 | `--lmcccm-link-text`                  | Link text color                                         |
 | `--lmcccm-link-hover-text`            | Link text color on hover                                |
 | `--lmcccm-link-active-text`           | Link text color in active state                         |
+| `--lmcccm-btn-border-width`           | Button border width                                     |
+| `--lmcccm-btn-border-style`           | Button border style                                     |
```
adamkudrna, line 321: I added these two properties so button borders can be adjusted.

PR #95, opened 2021-11-26
Docs: Improve docs design

adamkudrna:
image

### DIFF
--- a/examples/callbacks.html
+++ b/examples/callbacks.html
@@ -16,50 +16,62 @@
 
     <link rel="stylesheet" href="../dist/LmcCookieConsentManager.min.css"><!-- You will load this from CDN instead -->
 </head>
-<body style="padding-bottom: 420px">
+<body style="padding-bottom: 380px">
 
-<header class="container mt-4">
-    <h1>LMC Cookie Consent Manager</h1>
+<header class="container">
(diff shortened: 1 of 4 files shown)


examples/callbacks.html
```diff
@@ -16,50 +16,63 @@
...
 
-            <div class="row">
-                <div class="col-md-6 col-xs-12">
+        </div>
+    </div>
+    <div class="sticky-top pt-5" style="margin-top: -3rem">
+        <div class="position-absolute top-100 start-50 translate-middle">
+            <a href="" class="btn btn-lg btn-warning shadow-lg" onclick="removeCookieAndReload();" data-bs-toggle="tooltip" data-bs-html="true" title="Remove <code>lmc_ccm</code> cookie and reload">
+                <i class="bi bi-arrow-clockwise"></i>
+            </a>
```
OndraM, line 68: Please make the button more explicit. Add some text like "Reset cookie". Having just the "reload" icon is confusing - without hovering for the tooltip you can't know what is its purpose.

The tooltip should be kept however, it explains the implementation details, which is useful in these examples.

OndraM (changes requested):
Wow, that is quite nice redesign! 👍 

I just came across this one issue:

OndraM (approved):

PR #104, opened 2021-11-29
Feat: Push to dataLayer only of first accept #CCM-46 

OndraM:


literat (approved):

src/LmcCookieConsentManager.js
```diff
@@ -149,4 +141,19 @@ const LmcCookieConsentManager = (serviceName, args) => {
   return cookieConsent;
 };
 
+/**
+ * @param {Object} cookie
+ */
+function pushToDataLayer(cookie) {
```
adamkudrna, line 147: Just for purity (and reading experience), shouldn't this function be defined before it's used, i.e. somewhere at the top of this file?

If `function` among `const` is the problem, then arrow function syntax may be used: `const pushToDataLayer = (cookie) => {}`. (@literat Shouldn't it be a default required by ESlint, BTW…?)

OndraM: See https://github.com/lmc-eu/cookie-consent-manager/pull/71#discussion_r757039259
:)

literat: @adamkudrna Well, maybe I think it should, because I am used to it. But as I wrote earlier: this is the difference between function declaration (this) and function expression (arrow function). This definition can be and is hoisted.

PR #110, opened 2021-11-30
Docs: Add table of contents to readme, grammar fixes

OndraM:
I would add this despite the "auto-TOC" GitHub provides: 

obrazek

Its nice... But hidden. Better to have the links and quick overview what is in the README right in front of the eyes :eyes: .

### DIFF
--- a/examples/configuration.html
+++ b/examples/configuration.html
@@ -41,7 +41,7 @@
 
             <h1 class="display-1 mt-md-3">Extended Configuration</h1>
             <p class="lead mb-4 mb-md-5">
-                This is an example of extended configuration.
+                This is an example of an extended configuration.
             </p>
 
             <ul class="nav nav-pills justify-content-center mb-3">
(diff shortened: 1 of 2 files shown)


examples/configuration.html
```diff
@@ -71,11 +71,28 @@ <h1 class="display-1 mt-md-3">Extended Configuration</h1>
         <div class="row justify-content-lg-center">
             <div class="col col-lg-10 col-xl-8">
 
+                <h2 class="mt-md-3 mb-3">What is shown here</h2>
+
+                <ul>
```
adamkudrna, line 76: I know it's a bad DX, but… Could you add `class="mb-4"` to all new `p`s and `ul`s? I can move it later to CSS of the Spirit Bootstrap skin, but for now, utilities are necessary to keep it consistent. Thanks!

adamkudrna (approved):

examples/configuration.html
OndraM: Ok.

PR #120, opened 2021-12-02
Chore/update eslint

literat:
Use stable version of ESLint config and use additional insight as recommended in readme.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,7 +14,7 @@ module.exports = {
 
   ignorePatterns: ['node_modules', '!.*.js'],
 
-  extends: ['@lmc-eu/eslint-config-base', 'prettier'],
+  extends: ['@lmc-eu/eslint-config-base', '@lmc-eu/eslint-config-base/optional', 'prettier'],
 
   plugins: ['prettier', 'jest', '@typescript-eslint'],
 
--- a/examples/assets/functions.js
+++ b/examples/assets/functions.js
@@ -1,6 +1,8 @@
 /* eslint-disable no-unused-vars */
 /**
  * Global function to reset cookie for the examples only
+ *
+ * @returns {void}
  */
 function removeCookieAndReload() {
   document.cookie = `lmc_ccm=; Max-Age=0; path=/; domain=${window.location.hostname}`;
--- a/examples/functions.js
+++ b/examples/functions.js
@@ -1,6 +1,8 @@
 /* eslint-disable no-unused-vars */
 /**
  * Global function to reset cookie for the examples only
+ *
+ * @returns {void}
  */
 function removeCookieAndReload() {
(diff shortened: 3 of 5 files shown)


src/consentCollector.js
```diff
@@ -1,9 +1,10 @@
 /**
  * Submit information about consent level given by the user.
  *
- * @param {string} consentCollectorApiUrl
- * @param {Object} cookieConsent
- * @param {boolean} acceptedOnlyNecessary
+ * @param {string} consentCollectorApiUrl - URL of the consent collector API
```
OndraM, line 4: What is value of such comments? The variable name is already self-descriptive enough and this is redundant... Comments for param and return makes only sense when they add some value. They should not be required - otherwise they will force you to write sometimes useless comments... It is better to focus on descriptive variable name.

literat: Those comments are not valid JsDoc. In this case description is missing, see https://eslint.org/docs/rules/valid-jsdoc.

But this rule is deprecated as I found out in that docs. So I've used eslint-plugin-jsdoc instead and output is more disturbing than this.

```
/Users/literat/Work/lmc/cookie-consent-manager/src/consentCollector.js
   5:0  warning  Invalid JSDoc @param "cookieConsent" type "Object"; prefer: "object"       jsdoc/check-types
   6:0  warning  Missing JSDoc @param "acceptedOnlyNecessary" description                   jsdoc/require-param-description
  16:0  warning  Invalid JSDoc @param "cookieConsent" type "Object"; prefer: "object"       jsdoc/check-types
  18:0  warning  Invalid JSDoc @returns type "Object"; prefer: "object"                     jsdoc/check-types
  18:0  warning  Missing JSDoc @returns description                                         jsdoc/require-returns-description
  44:1  warning  Missing JSDoc @returns declaration                                         jsdoc/require-returns
  46:0  warning  Invalid JSDoc @param "payload" type "Object"; prefer: "object"             jsdoc/check-types
  47:0  warning  Invalid JSDoc tag (preference). Replace "return" JSDoc tag with "returns"  jsdoc/check-tag-names
```

literat: I also think that this case will be solved by #100 :-)

OndraM: Ok, but can't we for now just set `requireParamDescription": false` for the rule (or it could be added to the base package, because I don't think forcing this is a good practice), instead of temporarily adding such worthless comments? 🤔

literat: Fixed in https://github.com/lmc-eu/code-quality-tools/pull/32.

literat:
@OndraM I think this is ready to be merged. Most of the problems where fixed in CQT base. So here is only dep update and minor jsdoc fixes.

OndraM (approved):
👍

PR #121, opened 2021-12-02
Chore: Disable dependabot because of doing lot of noise

literat:
Limiting dependabot PRs to zero to disable it until we come with better solution.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -5,7 +5,7 @@ updates:
     schedule:
       interval: daily
       timezone: Europe/Prague
-    open-pull-requests-limit: 3
+    open-pull-requests-limit: 0
     versioning-strategy: increase-if-necessary
     commit-message:
       prefix: Deps


literat:
Please consider to keep automated dependencies live. I know that it is making a lot of noise but I think it is good to be up-to-date in project dependencies.

I think of it a lot and there are few solutions we can use or give them a try:
* Stay with Dependabot and switch form `daily` interval to `weekly` or `monthly`
* Stay with Dependabot and `ignore` or `allow` certain dependencies (https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#allow) - in our case are post important `dependencies` or simply `vanilla-cookieconset` because we want to be noticed with new updates to this core library
* Switch to RenovateBot which allows more detailed configuration like grouping or automerging (https://docs.renovatebot.com/configuration-options/), we can also apply some or all rules from above

Please, feel free to write any other suggestion or solution to keep us and our email clients sane.

OndraM (commented):
Maybe remove the whole dependabot file altogether? Once Renovate is setup, we don't need it, right?

(BTW maybe the Dependabot disable could be added as a commit to #122, and this PR could  be closed.)

literat:
Moved to #122

PR #142, opened 2021-12-06
Chore(renovate): Schedule dependency updates on weekly basis

literat:


renovate.json
```diff
@@ -12,6 +12,7 @@
   "branchPrefix": "dependencies/",
   "commitMessagePrefix": "Deps: ",
   "labels": ["dependencies"],
+  "schedule": ["after 10pm and before 7am on Monday"],
```
OndraM, line 15: This would mean doing the updates in timeframe 00:00-7:00 and 22:00-23:59 on Monday, wouldn't it?

I would suggest something like 9:00-22:00, ie. `after 9am and before 10pm on Monday`. So its during business hours.

literat: I think that business hours are `9am - 17pm`. I was thinking about it as I want that all updates should be prepared before I started my work and during the work it is quiet and I have the time to deal with it without interrupting with other updates.

OndraM: But this would mean you cannot finish the updates during a day -

OndraM: But this settings would mean you will receive work-related tasks to your e-mail during night, which is kinda rude... I think you should get those e-mails during office hours, not before and after work. And as this is work-related and not free-time project (where I'm ok with non-businesshours schedules :upside_down_face: ), I would incline to more daytime schedule. 9-17 is fine as well.

literat: Ok, updated.

renovate.json
```diff
@@ -12,6 +12,7 @@
   "branchPrefix": "dependencies/",
   "commitMessagePrefix": "Deps: ",
   "labels": ["dependencies"],
+  "schedule": ["after 9am and before 17pm on Monday"],
```
OndraM, line 15:
```suggestion
  "schedule": ["after 9am and before 5pm on Monday"],
```

(or if it can handle 24h format, then `after 9:00 and before 17:00` - but I didn't find anything about this in docs 🤔)

OndraM (changes requested):

OndraM (approved):

OndraM (approved):

renovate.json
literat: Oh shit, my bad 🤦‍♂️  My entire life is in 24h format :-)

PR #158, opened 2021-12-13
Readme improvements

OndraM:
Also the tables were reformatted by new PhpStorm... 🙃


literat (approved):

README.md
```diff
@@ -148,19 +148,22 @@ consent category. This must be done *before* the respective cookie is set.
...
+| `personalization` | For personalization based on user profiling (recommendation, etc.)                                                                                                                                                                                                                                                                                                                                                                                                                                                                          |
 
 ### GTM (Google Tag Manager) scripts
 
-GTM scripts are managed centrally, so after implementing the library you don't need to worry about them.
+GTM scripts are managed centrally in LMC, so after implementing the library you don't need to worry about them.
 The library sets all the required data to GTM dataLayer.
 
+However, keep in mind you still need to **take care of already existing cookies** (i.e. cookies created before
```
czhDavid, line 164: Is this correct? what about scenario where user gives us consent, then revokes it (we need to ask him again after some time. On apple devices even every two weeks). As far as I know we still need to clear those cookies no matter when the library was implemented. 
This is a bit confusing. 
Lets just state that we should delete even cookies that could be created by GTM and avoid all the confusion?

czhDavid (changes requested):

README.md
OndraM: Yes, you are right, sorry, I didn't realized that. 

So I remove the bracket contents to not confuse, have a look please.

czhDavid (approved):

PR #159, opened 2021-12-14
Chore: Annotate the CommonJS export names for ESM import in Node

literat:
* we need to build commonjs module as `browser` to avoid nanoid using
    crypto module which is native node.js module
  * instead of that we want to use web crypto api
  * using `platform:browser` bundler finds right export from nanoid and
    use it
  * but for this platform bundler did not annotate the commonjs export
    so we must do it manually

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -1,6 +1,6 @@
 /* eslint-disable no-console */
 const { build } = require('esbuild');
-const replace = require('replace-in-file');
+const fs = require('fs');
 
 // iife
 build({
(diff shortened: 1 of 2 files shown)


scripts/postbuild.js
```diff
@@ -0,0 +1,13 @@
+const fs = require('fs');
+
+// Annotate the CommonJS export names for ESM import in node
+fs.appendFile(
```
OndraM, line 4: Can't we do this in build.js instead?

There is already `replace` step which adds the `module.exports` (which will not not apply).
It could just be changed to this `fs.appendFile`, and we would not need this `postbuild.js`.

OndraM (changes requested):

scripts/postbuild.js
literat: 🤦

scripts/build.js
```diff
@@ -31,30 +31,24 @@ build({
...
-  .catch((error) => {
-    console.error(error);
-    process.exit(1);
-  });
+}).then(() => {
+  // Annotate the CommonJS export names for ESM import in node
+  fs.appendFile(
+    './dist/LmcCookieConsentManager.cjs',
+    `// Annotate the CommonJS export names for ESM import in node:
+module.exports = LmcCookieConsentManager_default`,
```
OndraM, line 46:
```suggestion
module.exports = LmcCookieConsentManager_default;
`,
```

(semicolon + newline on the end of the file)

OndraM (changes requested):

OndraM (approved):
👍 thanks

PR #164, opened 2021-12-15
Feat: Set cookie expiration when only necessary category accepted

OndraM:
See [CCM-53](https://jira.int.lmc.cz/browse/CCM-53) .

- [x] Blocked by https://github.com/orestbida/cookieconsent/issues/152
- [x] Make sure we send proper value to the API


literat (approved):

OndraM:
The exact expiration time is still under discussion, could be changed later.

OndraM:
> The exact expiration time is still under discussion, could be changed later.

=> Would be 60 by default, so this PR is up to date.

literat (approved):

PR #169, opened 2021-12-17
Chore: Trigger deploy preview only when certain directories changed

literat:
* using custom netlify build hook
  * @see https://docs.netlify.com/configure-builds/build-hooks/

### DIFF
--- a/bin/ci/deploy-preview.sh
+++ b/bin/ci/deploy-preview.sh
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
 
-$branch = $1
-
 function has_code_changes() {
   git diff --quiet HEAD origin/main -- src examples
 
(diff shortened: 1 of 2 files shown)


bin/ci/deploy-preview.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+function has_code_changes() {
+  git fetch --no-tags --depth=50 origin main
```
OndraM, line 4: Whats the purpose of this fetch? If we need depth > 1 (which is `actions/checkout@v2` default), we can use `fetch-depth: 50` right in the build step, see https://github.com/actions/checkout/

bin/ci/deploy-preview.sh
```diff
@@ -0,0 +1,25 @@
...
+    echo "Code changes detected."
+
+    return 1
+  fi
+}
+
+function trigger_build() {
+  echo "Triggering build..."
+
+  curl -X POST -d '{}' https://api.netlify.com/build_hooks/61bca39f3b89823193476ecd?trigger_branch=$1
```
OndraM, line 21: Don't you want to put the hook id to secrets? 🤔

OndraM (changes requested):
BTW, is the shell script even necessary? Cannot the same be accomplished right in action config? Like:

```yaml
on:
  push:
    paths:
      - 'src/'
      - 'examples/'
```

https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-including-paths

literat:
Closing this because we are deploying under the LMC account from now.

PR #170, opened 2021-12-17
Chore/commitlint

literat:


OndraM (approved):
Just an idea - what about moving this to `package.json`? https://github.com/conventional-changelog/commitlint#config

Less config files (especially when this is just a reference to parent config) may be even better.

literat:
> Just an idea - what about moving this to `package.json`? https://github.com/conventional-changelog/commitlint#config
> 
> Less config files (especially when this is just a reference to parent config) may be even better.

Good Idea! But All or Nothing! Eg. we can store all configs in `package.json` or none of them. If we store there only part of configs it could lead into inconsistency and misleading developers where to store or find the which config is used.

In a near future all configs should be short enough to place them into `package.json`.

OndraM:
> But All or Nothing!

You would say "Better something than nothing" :-D . The root scope is filled with so many config files, making it already a bit confusing so any cleanup would be welcomed to me, but whatever, it was just an idea :) . LGTM!

PR #173, opened 2021-12-21
Feat: Introduce `cookieName` option for setting cookie name

literat:


literat:
@OndraM Should it stay or should it go?

As we have an internal debate. Cookie name is configurable via `options.config` so this implementation is not necessary and not needed.

OndraM:
Agreed, closing this.

PR #181, opened 2022-01-11
Feat: Add Czech translations for settings modal

OndraM:
Also added a placeholder for other translations, to that this branch could be used as a base branch for styling of the component by @adamkudrna .


literat (approved):

src/languages/cs.ts
```diff
@@ -39,36 +40,61 @@ export const config = (extraMessages: ExtraMessages): VanillaCookieConsent.Langu
       },
     },
     settings_modal: {
+      title: 'Vlastní nastavení souborů Cookies',
```
literat, line 43: This changed in #183. So I expect this will be merged first.

PR #183, opened 2022-01-14
Feat: Add settings modal translations for all remaining languages

OndraM:


src/languages/de.ts
```diff
@@ -33,53 +33,57 @@ export const config = (extraMessages: ExtraMessages): VanillaCookieConsent.Langu
...
           toggle: {
             value: CookieConsentCategory.FUNCTIONALITY,
             enabled: false,
             readonly: false,
           },
         },
         {
-          title: 'AD TODO',
-          description: `Lorem ipsum`,
+          title: 'Marketing-Cookies',
```
literat, line 76: `Marketing-Cookies` -> `Marketing Cookies` ?

literat:
I have a problem with some wording because it sounds weird to me. Like in English `your own settings`. Or Czech or Slovak `vlastní nastavení`. There is nothing mine on a webpage unless I have access to some kind of an admin panel. So maybe `in the settings` is better, or `in the custom settings`.

We can use `Nastavení cookies` in Czech instead of `Vlastní nastavení` because I think it is not a correct translation of a `custom settings`, maybe `podrobné nastavení`. Maybe the same in other languages.

What do you think?

literat (changes requested):

OndraM:
> What do you think?

This would have to be discussed elsewhere - this is just an implementation of CCM-55 .

src/languages/de.ts
OndraM: Updated

OndraM:
@literat Translations updated.

literat (approved):

PR #199, opened 2022-01-28
Chore: Watch and build SCSS during development

adamkudrna:
Pulled out from PR #197.


package.json
```diff
@@ -69,7 +70,8 @@
     "postversion": "echo 'Check and push: `git push --set-upstream origin main && git push --tags`'",
     "release": "yarn version --`./bin/ci/semver.sh`",
     "replace-version": "ts-node scripts/readme-replace-version.ts",
-    "types": "tsc -p ./tsconfig.json"
+    "types": "tsc -p ./tsconfig.json",
+    "watch:css": "nodemon --watch src/scss/ --ext scss --exec \"yarn css\""
```
literat, line 74: I am expecting to call a `css:watch` script or `js:watch` script to watch for file change. It is the same as `unit:test:watch`. So why is this different?

adamkudrna: There is no reason, I just didn't notice there was another watch task. Fixed.

literat (approved):

OndraM (approved):

PR #200, opened 2022-02-02
Feat: Simplify translations #CCM-60 

OndraM:


src/scss/style/_typography.scss
```diff
@@ -12,6 +12,10 @@
     &::before {
         line-height: tokens.$line-height-default;
     }
+
+    strong {
```
OndraM, line 16: Needs to be verified by @adamkudrna :upside_down_face:

src/scss/style/_typography.scss
```diff
@@ -12,6 +12,10 @@
...
+        font-weight: tokens.$font-weight-bold;
+    }
```
OndraM, line 18: Needs to be verified by @adamkudrna :upside_down_face:

adamkudrna (approved):

PR #206, opened 2022-02-08
Docs: Rework callback examples #CCM-57

OndraM:


examples/callbacks.html
```diff
@@ -112,29 +124,32 @@ <h2 class="mt-md-3 mb-3">Callbacks</h2>
...
+        onFirstAccept: (cookieConsent) => { // user just clicked and gave consent
+          markCallbackCalled('onFirstAccept');
 
           if (cookieConsent.allowedCategory('functionality')) { // check for specific consent category
             console.log('Consent for extended functionality given');
           }
         },
-        onChange: (cookieConsent, categories) => { // consent change detected
+        onChange: (cookieConsent, categories) => { // user previously gave consent but now change its preferences
+          resetCallbackMarks();
```
OndraM, line 135: This is to not keep previous "onAccept" marked as "called". After uses updates the settings, only onChange is triggered, so the other callbacks should not be show as called - it would be quite confusing.

literat (approved):

PR #201, opened 2022-02-03
Feat: Tweak box mode of consent modal according to graphic design #CCM-47

adamkudrna:
obrazek

Snímek obrazovky 2022-02-03 v 18 54 26

PR #208, opened 2022-02-10
Feat: Add displayMode (default `force`) to choose between blocking pop-up and banner

OndraM:
With #201, the "box" layout for consent window is also possible.

However, enabling it is too complex, as one would have to override the whole gui_settings (the is no deep-merge when providing config) just to enable it:

```js
initLmcCookieConsentManager(
  'github.example',
  {
    config: {
      gui_options: {
        consent_modal: {
          layout: VanillaCookieConsent.GuiConsentLayout.BOX,
          position: VanillaCookieConsent.GuiConsentPosition.MIDDLE_CENTER, 
          transition: VanillaCookieConsent.Transition.SLIDE,
          swap_buttons: true,
        },
        settings_modal: {
          layout: VanillaCookieConsent.GuiSettingsLayout.BOX,
          transition: VanillaCookieConsent.Transition.SLIDE,
        },
      },
    }
  }
);
```

Also the box layout makes sense together with `force_consent: true` (when you want to force user action, you most probably want the consent to be shown in the middle of the page, as it is blocking user actions). On the other hand, when you want the consent to be less obtrusive, you probably want to show it in the bottom banner and with `force_consent: false`.

This is what `displayMode` does - it consistently applies those two settings (gui_options and force consent), to cover requirements of most if not all LMC products:
- `force` to show the consent in centered pop-up blocking modal
- `soft` to show the consent in non-blocking  bottom banner

`force` is now the new default, because it will be probably used by the vast majority of products.

This is kind-of-RFC, what do you think?

## TODO
- [x] Extend upgrade guide to cover using this option


literat (approved):

OndraM:
@literat Added upgrade guide

literat (changes requested):

literat:
> @literat Added upgrade guide, have a look please :)

@ondram Fix one sentence, please. And you are ready to merge :-)

PR #216, opened 2022-02-22
Feat: Allow to override default translation for consent modal title #CCM-62

OndraM:
This allows consent modal title translation to be overridden.

Overrides could be provided only for some languages.

Ability to override also other fields (button lables etc.) is expected in the future.


src/languages/cs.ts
```diff
@@ -18,7 +18,7 @@ export const config = (extraMessages: ExtraMessages): VanillaCookieConsent.Langu
 
   return {
     consent_modal: {
-      title: 'Díky Cookies budou naše stránky ještě lepší',
+      title: extraMessages.consentTitle ?? 'Díky Cookies budou naše stránky ještě lepší',
```
literat, line 21: I see a little problem here when using this null coalescing operator. When there will be values like `false` or `""` (empty string), this operator will not replace it. More often in this case is use `||` (OR operator) which replaces all bullish values.

I do not know if this was your intention. In most scenarios, this will work just fine but there are those edge cases I've mention above.

src/LmcCookieConsentManager.ts
```diff
@@ -51,6 +52,7 @@ const defaultOptions: CookieConsentManagerOptions = {
  *   to be given to multiple companies.
  * @param {DisplayMode} [args.displayMode] - `force` to show consent in a centered modal box and to block page until
  *   user action. `soft` to show consent in a banner on the bottom of the page.
+ * @param {Record<String, TranslationOverride>} [args.translationOverrides] - Translation overrides for specified languages
```
literat, line 55: You should define a new type instead of `String`. This mean now I can set for instance
```js
{ giberish: { consentTitle: 'some title' } }
```
and for type checking it will be fine.

Please create an union type `type Languages = 'cs' | 'sk' | ...` and replace the `String`.

src/LmcCookieConsentManager.ts
```diff
@@ -71,20 +73,21 @@ const LmcCookieConsentManager: CookieConsentManager = (serviceName, args) => {
...
   const languages = {
-    cs: configCs({ companyNames }),
-    de: configDe({ companyNames }),
-    en: configEn({ companyNames }),
-    hu: configHu({ companyNames }),
-    pl: configPl({ companyNames }),
-    ru: configRu({ companyNames }),
-    sk: configSk({ companyNames }),
-    uk: configUk({ companyNames }),
+    cs: configCs({ companyNames, consentTitle: translationOverrides.cs?.consentTitle }),
```
literat, line 83: It is only an idea but maybe we should be consistent with naming in the original language config. Seem to me to be better when I want to override `title`, I will use `title` key in the override object. I know that this is for now unique case and we do not want to support an entire language override but consider it, please.

literat (changes requested):

src/LmcCookieConsentManager.ts
OndraM: But there is `title` key for `consent_modal` and also `title` key for `settings_modal`, so using just `title` is not specific enough. I think we should make the name unique, even when we for now override only the modal title - as thepossibility of overriding also settings modal title in the future is real :).

src/languages/cs.ts
OndraM: You're right, but I thought of eg. `""` as an intentional value - if you want to not provide any title at all, for some A/B test for instance. So this behavior is IMO fine.

src/LmcCookieConsentManager.ts
OndraM: I thought about it, but we already use this type-free string for defaultLang.

I'm fine with adding the union type on both places, though. The only issue whit this is if someone would want to override the language via config and add some new language (which we don't include in the default), this type check will not allow to use the defaultLang and the translationOverrides. However, this is maybe just a theoretical scenario, what do you think?

src/LmcCookieConsentManager.ts
literat: Ok.

src/languages/cs.ts
literat: Ok.

literat (approved):

PR #218, opened 2022-03-03
Fix/missing exported types

literat:
Some types were missing when using extended configuration on Jobs.cz.

### DIFF
--- a/src/types/vanilla-cookieconsent.ts
+++ b/src/types/vanilla-cookieconsent.ts
@@ -159,6 +159,6 @@ export namespace VanillaCookieConsent {
     onAccept?: () => void;
     onChange?: () => void;
     onFirstAction?: () => void;
-    languages?: Languages;
+    languages?: Record<string, Languages>;
   }
 }
--- a/examples/webpack-with-cjs/index.js
+++ b/examples/webpack-with-cjs/index.js
@@ -1,6 +1,18 @@
 // eslint-disable-next-line import/no-unresolved
 const LmcCookieConsentManager = require('@lmc-eu/cookie-consent-manager');
+// eslint-disable-next-line import/no-unresolved
+const { DisplayMode } = require('@lmc-eu/cookie-consent-manager');
 
-window.addEventListener('load', () => {
-  LmcCookieConsentManager('example');
+const ccmArgs = {
+  displayMode: DisplayMode.SOFT,
+  config: {
+      cookie_name: 'lmc_ccm_example',
+      force_consent: true,
+      cookie_domain: 'example.com',
+      cookie_necessary_only_expiration: 30,
+  }
+};
+
+window.addEventListener('DOMContentLoaded', () => {
+  LmcCookieConsentManager('example', ccmArgs);
 });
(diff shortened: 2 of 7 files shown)


D0L1K:
Please add `ExtraMessages` from types, `addSeparators` and `pluralize` from utils to export

OndraM (approved):
There is a typo in the commit message: `Docs: Use concent configuration in examples` - please don't forget to fix it when rebasing.

OndraM:
@D0L1K 
Wouldn't #216 and https://jira.int.lmc.cz/browse/CCM-63 solve the issue for you? We are making parts of the translations overridable, so maybe you wouldn't need to define your own translations, just override what you need.

D0L1K:
> @D0L1K Wouldn't #216 and https://jira.int.lmc.cz/browse/CCM-63 solve the issue for you? We are making parts of the translations overridable, so maybe you wouldn't need to define your own translations, just override what you need.

Well I only need to change the title for cs on Jobs.cz and Prace.cz, but my last information (and as far as I understand the code) only way now is to specify everything in the config.languages, that's why I've started with copying your code. But I've already change it to be able to specify it without those internal functions, so that's not the problem. And once you will solve changing specific translation parts or at least the title (https://jira.int.lmc.cz/projects/CCM/issues/CCM-62), it will make my case much more easier :)

OndraM:
> Well I only need to change the title for cs on Jobs.cz and Prace.cz, but my last information (and as far as I understand the code) only way now is to specify everything in the config.languages, that's why I've started with copying your code. 

Since 2.0 you can get current translations using eg. `window.ccm.getConfig('languages');`, but generally yes, there is no simple way.

> And once you will solve changing specific translation parts or at least the title (https://jira.int.lmc.cz/projects/CCM/issues/CCM-62), it will make my case much more easier :)

Right, CCM-62 is implemented in #216 , should be part of release 2.1 soon :) .

literat:
> Right, CCM-62 is implemented in https://github.com/lmc-eu/cookie-consent-manager/pull/216 , should be part of release 2.1 soon :) .

After study of the original PR on the product, this should be enough to solve the needs of the product.

literat:
> There is a typo in the commit message: `Docs: Use concent configuration in examples` - please don't forget to fix it when rebasing.

Fixed, thanks for notice.

src/types/vanilla-cookieconsent.ts
```diff
@@ -159,6 +159,6 @@ export namespace VanillaCookieConsent {
     onAccept?: () => void;
     onChange?: () => void;
     onFirstAction?: () => void;
-    languages?: Languages;
+    languages?: Record<string, Languages>;
```
OndraM, line 162: 🤔 How is that eslint didn't complain?

literat: Because `vanilla-cookieconsent` lack of type definitions. So here https://github.com/lmc-eu/cookie-consent-manager/blob/main/src/LmcCookieConsentManager.ts#L173 cannot be done any type check when using options.

I should make it compatible and safe by adding `VanillaCookieConsent.Options` to the https://github.com/lmc-eu/cookie-consent-manager/blob/main/src/LmcCookieConsentManager.ts#L135. But it will be needed to recheck and define all missing options. Maybe in another PR.

OndraM (approved):

PR #220, opened 2022-03-04
Allow overriding consent description intro #CCM-63

OndraM:


src/utils.ts
```diff
@@ -12,3 +12,9 @@ export const addSeparators = (strings: string[], and: string = ''): string =>
   });
 
 export const pluralize = (count: number, singular: string, plural: string): string => (count === 1 ? singular : plural);
+
+export const assembleDescriptionIntro = (defaultValue: string, overrideValue?: string): string => {
```
literat, line 16: I am missing a test for this function. Could you please add it?

literat (changes requested):
Please, add a missing test.

src/utils.ts
OndraM: Did you miss this? 

https://github.com/lmc-eu/cookie-consent-manager/pull/220/files#diff-3c616be935d4f3f12995ccdc0d101840c0bb76665a29c61a090a71ac9be6b6b6R34

literat (approved):

PR #222, opened 2022-03-07
Fix: Provide type for vanilla cookie consent options

literat:
* ensure that vanilla cookie consent configuration is type safe
  * @see https://github.com/lmc-eu/cookie-consent-manager/pull/218#discussion_r818638417


OndraM (approved):
Except for the conflict LGTM :)

PR #228, opened 2022-03-21
Fix/isolated modules

literat:
* `export const enum` do not work well with `isolatedModules` options
* use rather object with named values
* we should do not export const enums ever in any library (for
   internal use it works well)
* @see: https://www.typescriptlang.org/tsconfig#isolatedModules
* @see: https://ncjamieson.com/dont-export-const-enums/
* @see: https://fettblog.eu/tidy-typescript-avoid-enums/

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,4 +1,4 @@
-name: Test
+name: Test build
 
 on:
   push:
@@ -6,10 +6,13 @@ on:
     - cron: '0 9 * * *' # every day at 9:00
 
 jobs:
-  test:
-    name: Tests and linting
+  test-build:
+    name: Build example
     runs-on: ubuntu-20.04
     timeout-minutes: 5
+    strategy:
+      matrix:
+        example: [webpack-with-cjs, webpack-with-esm, webpack-with-typescript]
 
     steps:
       - name: Checkout
@@ -26,14 +29,8 @@ jobs:
       - name: Install dependencies
         uses: bahmutov/npm-install@v1
 
-      - name: Lint
-        run: yarn lint
+      - name: Test build
+        run: yarn build
 
-      - name: Check code-style format
-        run: yarn format
-
-      - name: Check types
-        run: yarn types
-
-      - name: Unit tests
-        run: yarn test:unit
+      - name: Test example build
+        run: cd ./examples/${{ matrix.example }} && yarn && yarn build
--- a/examples/webpack-with-typescript/index.ts
+++ b/examples/webpack-with-typescript/index.ts
@@ -1,0 +1,15 @@
+// eslint-disable-next-line import/no-unresolved
+import LmcCookieConsentManager, { DisplayMode } from '@lmc-eu/cookie-consent-manager';
+
+const ccmArgs = {
+  displayMode: DisplayMode.SOFT,
+  config: {
+      cookie_name: 'lmc_ccm_example',
+      cookie_domain: 'example.com',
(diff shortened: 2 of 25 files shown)


src/types/CookieConsentManager.ts
```diff
@@ -1,24 +1,25 @@
-import { CookieConsentCategory } from './CookieConsentCategory';
+import { CookieConsentCategory, DisplayMode } from '../constants';
 import { VanillaCookieConsent } from './vanilla-cookieconsent';
 
+export type Values<T> = T[keyof T];
+
+export type CookieConsentCategoryValues = Values<typeof CookieConsentCategory>;
+
```
OndraM, line 7: The purpose of this is to workaround for not using enums, right?

So basically we will have to take care of variables which are basically "enums" in a two different ways, depending on whether the type is part of exported public API (so you must use generics like this, ie. emulate enum behavior) or whether they are used only internally (when you use "real" enums")? That is quite a confusing... 🤔

literat: Well, not so far confusing. This is not the first time I encounter some problem with enums. For example iterating over enum and getting values from them is a task from hell (search something like "how to use enum as index key type in typescript").

So my opinion is to ban enums for type safety reasons. Not just that we are developing a library but also for sake of our souls. It is better not to think about whether I am using enum internally or not. Just do not use them everywhere.

literat:
It has been confirmed that this is working solution.

examples/webpack-with-typescript/index.ts
```diff
@@ -0,0 +1,16 @@
+// eslint-disable-next-line import/no-unresolved
+import LmcCookieConsentManager, { DisplayMode } from '@lmc-eu/cookie-consent-manager';
+
+const ccmArgs = {
+  displayMode: DisplayMode.SOFT,
+  config: {
+      cookie_name: 'lmc_ccm_example',
+      force_consent: true,
```
OndraM, line 8: Please remove overriding force_consent from the examples... There is already the displayMode being used, so either use it as soft or force, but don't combine it with force_consent in the examples. :)

OndraM (changes requested):

src/types/CookieConsentManager.ts
```diff
@@ -1,24 +1,25 @@
-import { CookieConsentCategory } from './CookieConsentCategory';
+import { CookieConsentCategory, DisplayMode } from '../constants';
 import { VanillaCookieConsent } from './vanilla-cookieconsent';
 
+export type Values<T> = T[keyof T];
```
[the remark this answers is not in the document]
literat: Here it is exported for internal use only. You must look at what is exported from the package main entry point which is `src/LmcCookieConsentManager.ts`

.github/workflows/build.yaml
```diff
@@ -0,0 +1,36 @@
+name: Build
```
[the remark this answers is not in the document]
literat: Renamed, sounds reasonable.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+on:
+  push:
+  schedule:
+    - cron: '0 9 * * *' # every day at 9:00
+
+jobs:
+  test:
+    name: Build
```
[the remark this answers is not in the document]
literat: Renamed.

examples/webpack-with-typescript/index.ts
literat: Removed.

OndraM (approved):

PR #234, opened 2022-04-19
Feat: Add secondaryButtonMode to choose secondary button behavior

OndraM:
#CCM-64


src/languages/cs.ts
```diff
@@ -30,17 +34,17 @@ export const config = (extraMessages: ExtraMessages): VanillaCookieConsent.Langu
         ${pluralize(lang.companyNames.length, lang.company, lang.companies)}
         ${addSeparators(lang.companyNames, lang.and)}
         souhlas s využíváním souborů Cookies na účely personalizace, analýzy a cíleného marketingu.
-        Další informace o Cookies a úpravu jejich používání naleznete
-        ve <strong><a href="" data-cc="c-settings">vlastním nastavení</a></strong>.
+        ${
+          secondaryButtonMode !== SecondaryButtonMode.SHOW_SETTINGS
```
literat, line 38: I think it will improve readability if there will be a named function like `isSecondaryButtonShown(secondaryButtonMode)` or something similar. This condition is the same in multiple files, so we can provide a function.

literat (changes requested):

src/LmcCookieConsentManager.ts
```diff
@@ -73,21 +77,22 @@ const LmcCookieConsentManager: CookieConsentManager = (serviceName, args) => {
...
     companyNames,
     displayMode,
+    secondaryButtonMode,
     translationOverrides,
     config,
   } = options;
   const cookieName = 'lmc_ccm';
   const cookieConsent = window.initCookieConsent();
 
   const languages = {
```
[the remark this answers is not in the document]
OndraM: I was unable to make this work, it seems not to be easy to call default export of a module by a string value.

For example this fails on `Uncaught TypeError: configCallback is not a function`.

src/languages/cs.ts
OndraM: Well, the logic is that this sentence should be added only if the "Open settings" button is not already shown as the secondary button. So its hard to find accurate name. (`isSecondaryButtonShown` would be misleading, there is always secondary button). What about `isSettingsButtonNotShown`?

src/LmcCookieConsentManager.ts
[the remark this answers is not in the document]
literat: Never mind, it was only suggestion, maybe ```window[`config${capitalize(language)}`]()``` should work.

src/languages/cs.ts
OndraM: Ok, updated, have a look please.

literat (approved):

PR #254, opened 2022-07-25
Docs: Fix typo in changelog message

literat:
`csent` -> `consent`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 
 ### Dependencies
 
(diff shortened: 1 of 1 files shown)


OndraM (approved):

OndraM:
> So some problem in markdown-link-check?

Github URLs are [intentionally disabled](https://github.com/lmc-eu/cookie-consent-manager/blob/main/.github/markdown-link-check-config.json), because GitHub blocks with HTTP 403 its thinking its DoS.

literat:
> Some github URLs are [intentionally disabled](https://github.com/lmc-eu/cookie-consent-manager/blob/main/.github/markdown-link-check-config.json), because GitHub blocks with HTTP 429 its thinking its DoS.

@OndraM Is there a way to solve this? Because now we have failing check in every PR.

literat:
> @OndraM Is there a way to solve this? Because now we have failing check in every PR.

The same link worked a few hours before: https://github.com/lmc-eu/cookie-consent-manager/runs/7496777997?check_suite_focus=true#step:4:508

Seems like a some kind of issue on GitHub.

PR #255, opened 2022-07-25
Docs: Remove integrity attributes because of integrity fails

literat:
* @see: https://www.jsdelivr.com/using-sri-with-dynamic-files
      * we are targeting on latest version 5 which means it changes
        dynamically and thus integrity attribute hash works until
        the next update
      * it is recommended to use them only with full version path


OndraM:
👍 This was probably forgotten when we changed the full versions to version aliases.

OndraM (approved):
👍 This was probably forgotten when we changed the full versions to version aliases.

PR #271, opened 2022-12-06
Replace LMC company name with translated legal name

OndraM:
See attached [JIRA issue](https://jira.lmc.cz/browse/CCM-65) for details.


literat (approved):
Reformat your code using `yarn format:fix` and the job is done :-)

PR #272, opened 2022-12-06
Deps: Update dependencies

OndraM:
This manually updates "safe" dependencies for current CCM version 2 (i.e. those which don't break anything and don't require further changes :) ) . 

(Context: We do have renovate bot disabled for the "main" branch, because its no longer regularly developed and focus was supposed to be moved to v3 branch, but its yet to happen. So we are in a kind of grey transition period, and I want to prepare the codebase for upcoming release which will include #271 .)

@adamkudrna Just please make sure there is no BC in spirit 0.17 -> 0.23.


adamkudrna (approved):
[No important changes](https://github.com/lmc-eu/spirit-design-system/compare/@lmc-eu/spirit-design-tokens@0.17.0...@lmc-eu/spirit-design-tokens@0.23.0#diff-1907fcdc512adfa452cfed058bf4684f288094fcbc7ba9bea353dffa474ffd7c) here, I also tested it manually and everything looks OK.

PR #301, opened 2023-06-12
Chore: Remove @adamkudrna and @crispeen from codeowners

literat:
I think there is no need to have these two super ninja developers on each PR of this project ;-)


OndraM (approved):
:ninja:

PR #314, opened 2023-11-02
Feat: Replace LMC company name with Alma Career

OndraM:
This changes "LMC" in the legal text with "Alma Career".

Note: If you override `companyNames` settings array, you must explicitly change `LMC` to `Alma Career`. If you use default settings, no action is needed for the change.


src/__tests__/utils.test.ts
```diff
@@ -3,7 +3,7 @@ import {
   assembleDescriptionIntro,
   assembleSecondaryButton,
   isSettingsButtonNotShown,
-  legalizeLmc,
+  legalizeAlmaCareer,
```
literat, line 6: What about renaming this to a more general name? I know that there is no expected company renaming in the near future, but 🤷

literat (approved):

src/__tests__/utils.test.ts
OndraM: But the method is not general - it does legalize this and only this name :) . So this IMHO describes it correctly.

PR #318, opened 2024-01-02
Docs: Rebrand to Alma Career also in documentation and meta information

OndraM:


src/scss/theme/_default.scss
```diff
@@ -1,5 +1,5 @@
 //
-// Read from LMC CCM public theming API and assign to private theming layer.
+// Read from Alma Career CCM public theming API and assign to private theming layer.
 //
 // `--lmcccm-p-*` (`p` for private) are internally used custom properties that either obtain
```
crishpeen, line 4: --accccm-p- 😬

crishpeen (approved):

LICENSE.md
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2021–2022 LMC s.r.o.
+Copyright (c) 2021–2024 Alma Career Czechia s.r.o.
```
literat, line 3: You will be changing this every year? 🤔

OndraM: No, I hope we won't rename every year! 🙃

literat: I mean the year number, e. g. `2024`. In 2025 this will be outdated, though.

OndraM: I know, but I don't understand what do you suggest? 2024 is the first year of Alma Career, so keeping the old year doesn't not make sense, as does not keeping the old company name. 🤔

literat: According [this document](https://www.copyrightlaws.com/copyright-symbol-notice-year/#:~:text=Which%20Year%20Should%20You%20Use,to%20the%20public%20without%20restriction.) the year of first publication should be enough. Although there is also a paragraph about blogs and websites where the range can be used. And the second number describes the year of the latest update. Maybe it makes sense but we should not forget to update the year everytime we make a change :D

OndraM: But 2024 is the first year of publication in the name of Alma Career. I don't plan updating this next year.

Maybe comma or another line would be better, but I don't think this is a big issue 😅

literat: I agree. No big deal. Just wonder 😅

literat (approved):

src/scss/theme/_default.scss
adamkudrna: We can be lucky it's not "Alma Career **Czechia**"

examples/assets/examples.css
```diff
@@ -1,5 +1,5 @@
 /*!
- * LMC Cookie Consent Manager
+ * Alma Career Cookie Consent Manager
```
adamkudrna, line 2: No "Czechia" here? I'm not sure but "Alma Media" alone is no subject…

PR #324, opened 2024-02-27
Feat: Add ability to set cookie table

OndraM:
- [x] add missing Slovenian translation

### DIFF
--- a/src/languages/__tests__/langConfig.test.ts
+++ b/src/languages/__tests__/langConfig.test.ts
@@ -24,11 +24,40 @@ describe.each([
 ])('config %s', (name, config) => {
   it('should return localization object for acceptNecessary button mode', () => {
     expect(
-      config({ companyNames: ['test1', 'test2', 'test3'] }, SecondaryButtonMode.ACCEPT_NECESSARY),
+      config({ companyNames: ['test1', 'test2', 'test3'] }, SecondaryButtonMode.ACCEPT_NECESSARY, {}),
     ).toMatchSnapshot();
   });
 
   it('should return localization object for showSettings button mode', () => {
-    expect(config({ companyNames: ['test1', 'test2', 'test3'] }, SecondaryButtonMode.SHOW_SETTINGS)).toMatchSnapshot();
+    expect(
+      config({ companyNames: ['test1', 'test2', 'test3'] }, SecondaryButtonMode.SHOW_SETTINGS, {}),
+    ).toMatchSnapshot();
+  });
+
+  it('should return localization object with cookie table', () => {
+    const cookieTable = {
+      necessary: [
+        {
+          name: '^_ga',
+          description: 'Google Analytics description ...',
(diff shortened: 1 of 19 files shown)


src/languages/sk.ts
```diff
@@ -73,6 +76,7 @@ export const config = (
             enabled: true,
             readonly: true,
           },
+          cookie_table: cookieTable[CookieConsentCategory.NECESSARY] || undefined,
```
OndraM, line 79: Why `undefined`? Because any other value will cause Orest library to render the table header, even if the table is defined as empty.

We could construct this whole object differently (and do not put the `cookie_table` key here at all, if cookieTable is empty at), but IMO its is not worth the effort...

literat: What about the function that will apply this configuration to the translations based on some config? It can iterate over existing categories. It is still not worth the effort?

src/languages/__tests__/langConfig.test.ts
```diff
@@ -24,11 +24,35 @@ describe.each([
...
+    expect(
+      config({ companyNames: ['test1', 'test2', 'test3'] }, SecondaryButtonMode.SHOW_SETTINGS, {}),
+    ).toMatchSnapshot();
+  });
+
+  it('should return localization object with cookie table', () => {
+    const cookieTable = {
+      necessary: [
+        {
+          col_pattern: '^_ga',
```
literat, line 41: I presume that the snake case is required by the Orest library...?

OndraM: No, its just arbitrary column identifier, it could be anything. This was from orest's example, but I changed it to something more readable, without a prefix.

literat: Ok, my question was targeting used casing and whether the snake-case is required by some Orest functionality or not. If so, I suggest using camelCase as we are using it almost everywhere in the code.

OndraM: All the keys are now single worded:

```
{ name: '_ga', title: 'Google Analytics', description: 'Popisek ...', expiration: '1 rok' },
```

src/languages/sk.ts
OndraM: Hmm, but I don't think its worth it right now... Or at least not inside this PR. We plan to add more languages in near future, so I will reconsider it then (or in the meantime, but not as part of this PR).  

For example to encapsulate it in  something like:
```js
blocks: [
    assembleCategory(CookieConsentCategory.ANALYTICS, 'Analytické cookies', 'Pomocou nich sledujeme, koľko ľudí náš web navštevuje a ako ho používajú. Vďaka tomu môžeme stránky a ďalšie služby neustále vylepšovať.'),
    ...
```

literat: I am ok with that. It was just an interesting topic from the beginning :-)

PR #325, opened 2024-03-12
Refactor block assembly

OndraM:
As discussed in #324, I made at least some DRY simplification of block assembly, so that the language files contains less logic. 

This PR will go before #324 , ie. #324 will be refactored to extend and utilize the `assembleCategoryBlock` method.

What do you think, @literat?

### DIFF
--- a/src/languages/cs.ts
+++ b/src/languages/cs.ts
@@ -5,6 +5,11 @@ import {
   isSettingsButtonNotShown,
   pluralize,
   legalizeAlmaCareer,
+  assembleCategoryNecessary,
+  assembleCategoryAnalytics,
+  assembleCategoryFunctionality,
+  assembleCategoryAd,
+  assembleCategoryPersonalization,
 } from '../utils';
 import { ExtraMessages, Values, VanillaCookieConsent } from '../types';
 import { CookieConsentCategory, SecondaryButtonMode } from '../constants';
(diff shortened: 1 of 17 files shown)


src/utils.ts
```diff
@@ -49,6 +49,19 @@ export const assembleSecondaryButton = (
...
 };
 
+export const assembleCategoryBlock = (
+  category: CookieConsentCategoryValues,
+  title: string,
+  description: string,
+  readonly: boolean,
+): VanillaCookieConsent.ModalBlock => {
+  return {
+    title: title,
```
literat, line 59:
```suggestion
    title,
```

You can use a shorthand when the key and variable name are the same name.

src/languages/cs.ts
```diff
@@ -65,56 +66,36 @@ export const config = (
...
-          title: 'Personalizační Cookies',
-          description: `Naše služby fungují lépe, když je můžeme přizpůsobit na míru konkrétnímu uživateli.
-          Povolením Personalizačních cookies zvýšíte šanci, že najdete právě takový obsah, jaký hledáte.`,
-          toggle: {
-            value: CookieConsentCategory.PERSONALIZATION,
-            enabled: false,
-            readonly: false,
-          },
-        },
+        assembleCategoryBlock(
```
literat, line 69: I have an idea, maybe it is optimized too much so please decide for yourself.

```
assembleNecessaryCategory(....),

const assembleNecessaryCategory = (...) => assembleCategoryBlock(CookieConsentCategory.NECESSARY, ...)
```

I know, four more functions, but you can take care only of translations and not setting the category.

literat (approved):
LGTM 👍

src/languages/cs.ts
OndraM: :thinking: Good idea, I though about something similar but abandoned the idea. But new day, new view, so why not, I like it now :)

src/utils.ts
```diff
@@ -52,3 +51,39 @@ export const assembleSecondaryButton = (
...
+const assembleCategoryBlock = (
+  category: CookieConsentCategoryValues,
+  title: string,
+  description: string,
+  readonly: boolean,
+): VanillaCookieConsent.ModalBlock => {
+  return {
+    title,
+    description,
+    toggle: { value: category, enabled: readonly, readonly: readonly },
```
literat, line 87:
```suggestion
    toggle: { value: category, enabled: readonly, readonly },
```

PR #326, opened 2024-03-26
Add baltic languages

OndraM:


README.md
```diff
@@ -290,13 +290,12 @@ initLmcCookieConsentManager( // when loaded as a module, these options are passe
...
 | `translationOverrides`   | Record<string, TranslationOverride> | `{}`                                                        | Override default translation for specified languages. `consentTitle`, `descriptionIntro` and `settingsModalMoreInfo` could be overridden.<br>[See example][examples-configuration]                            |
-| `cookieTable`            | CookieTable                         | `{}`                                                        | Define cookie table for specified languages and specified categories<br>[See above for example configuration](#configuration)                                                                                 |
+| `cookieTable`            | CookieTable                         | `{}`                                                        | Define cookie table (which is shown in settings modal) for specified languages and specified categories.<br>[See above for example configuration](#configuration)                                             |
```
literat, line 293:
```suggestion
| `cookieTable`            | CookieTable                         | `{}`                                                        | Define the cookie table (shown in the settings modal) for specified languages and categories.<br>[See above for example configuration](#configuration)                                             |
```

README.md
```diff
@@ -290,13 +290,12 @@ initLmcCookieConsentManager( // when loaded as a module, these options are passe
...
 
 ### Supported languages
 
 Translation of the user interface is provided in the following languages:
-Czech (`cs`), German (`de`), English (`en`), Croatian (`hr`), Hungarian (`hu`), Polish (`pl`), Russian (`ru`),
-Slovak (`sk`), Slovenian (`sl`) and Ukrainian (`uk`).
+Czech (`cs`), German (`de`), English (`en`), Estonian (`et`), Croatian (`hr`), Hungarian (`hu`), Lithuanian (`lt`), Latvian (`lv`), Polish (`pl`), Russian (`ru`), Slovak (`sk`), Slovenian (`sl`) and Ukrainian (`uk`).
```
literat, line 298: Maybe it would be more convenient and readable if you used a list instead of a sentence.

src/LmcCookieConsentManager.ts
```diff
@@ -87,19 +90,29 @@ const LmcCookieConsentManager: CookieConsentManager = (serviceName, args) => {
...
-    de: configDe({ companyNames, ...translationOverrides.de }, secondaryButtonMode, cookieTable.de || {}),
-    en: configEn({ companyNames, ...translationOverrides.en }, secondaryButtonMode, cookieTable.en || {}),
-    hr: configHr({ companyNames, ...translationOverrides.hr }, secondaryButtonMode, cookieTable.hr || {}),
-    hu: configHu({ companyNames, ...translationOverrides.hu }, secondaryButtonMode, cookieTable.hu || {}),
-    pl: configPl({ companyNames, ...translationOverrides.pl }, secondaryButtonMode, cookieTable.pl || {}),
-    ru: configRu({ companyNames, ...translationOverrides.ru }, secondaryButtonMode, cookieTable.ru || {}),
-    sk: configSk({ companyNames, ...translationOverrides.sk }, secondaryButtonMode, cookieTable.sk || {}),
-    sl: configSl({ companyNames, ...translationOverrides.sl }, secondaryButtonMode, cookieTable.sl || {}),
-    uk: configUk({ companyNames, ...translationOverrides.uk }, secondaryButtonMode, cookieTable.uk || {}),
+  const languageConfigs = {
```
literat, line 93: As this is growing a little bit, I suggest moving this to a new file, maybe `languages/loader.ts` or something like that.

There you should import all those languages and here just use some function like `const languages = loadLanguages(translationOverrides);` because I think we are mixing here too much logic which can be separated for better maintainability.

Also, try to consider, if the dynamic import can be used based on file name, e. g. automagically import all files in languages based on time in some loop.

src/LmcCookieConsentManager.ts
```diff
@@ -87,19 +90,29 @@ const LmcCookieConsentManager: CookieConsentManager = (serviceName, args) => {
...
+    hu: configHu,
+    lt: configLt,
+    lv: configLv,
+    pl: configPl,
+    ru: configRu,
+    sk: configSk,
+    sl: configSl,
+    uk: configUk,
   };
+  const languages = Object.entries(languageConfigs).reduce((acc, [code, configFunction]) => {
```
literat, line 108: Please, consider creating a function like `normalizeLanguages` or `applyLangOverrides` for better readability. The reduce function here is a great and powerful choice but it lacks readability when not used in the named function, e.g. you must read entire reduce to understand what does it do instead of just reading the function name and then check if the code is correct.

src/LmcCookieConsentManager.ts
```diff
@@ -87,19 +90,29 @@ const LmcCookieConsentManager: CookieConsentManager = (serviceName, args) => {
...
+    acc[code] = configFunction(
+      { companyNames, ...translationOverrides[code] },
+      secondaryButtonMode,
+      cookieTable[code] || {},
+    );
+    return acc;
```
literat, line 114: Please, consider adding a free line before `return` for better readability. Also, I suggest using a better name for accumulated values, e.g. what is in the variable.
```suggestion

    return overridedTranslations;
```

literat (commented):
I did not try to check if the translations are correct :-) I have written down a few suggestions you should take care of.
Otherwise, it is looking good and working.

IMHO, we should think about a better solution for translations. Mixing translation string and business logic in this scale is not good for maintainability. 🤔

README.md
OndraM: I will probably do this in next PR, we will be adding even more languages...

src/LmcCookieConsentManager.ts
OndraM: > automagically import all files in languages based on time in some loop.

A considered and even tried this, but I don't like it because the magic imports are then not detectable by IDEs, and it will complain about unused imports, or you cannot easily find usages etc., so I prefer the declaration to be explicit.

As for the other suggestion (loader for languages) - you're right, it has grown a bit too much, I will probably do something like this in next PR, thanks for the suggestions.

OndraM:
> IMHO, we should think about a better solution for translations. Mixing translation string and business logic in this scale is not good for maintainability. 🤔

:+1: Yes, agreed, we actually recently discussed such plans to move translations to some more standard solution (like key-value tuples json) , and we may need to do it in some point later. Because current solution is hard to maintain no only when adding new translations, but in case of adding new/modifying current  translation strings.

src/LmcCookieConsentManager.ts
```diff
@@ -146,6 +151,17 @@ const LmcCookieConsentManager: CookieConsentManager = (serviceName, args) => {
     onChange(cookieConsent, categories);
   };
 
+  const assembleLanguagesConfig = () => {
+    return Object.entries(languagesMap).reduce((languagesConfig, [code, configFunction]) => {
```
literat, line 155: You can use an implicit return here in the arrow function `=> Object.entries`.

src/LmcCookieConsentManager.ts
```diff
@@ -146,6 +151,17 @@ const LmcCookieConsentManager: CookieConsentManager = (serviceName, args) => {
...
+      languagesConfig[code] = configFunction(
+        { companyNames, ...translationOverrides[code] },
+        secondaryButtonMode,
+        cookieTable[code] || {},
+      );
+      return languagesConfig;
```
literat, line 161: There has been a suggestion to add an empty line before this `return`.

literat (approved):
Minor things, but looking good 👍

PR #328, opened 2024-04-04
Add Macedonian and Bosnian translations

OndraM:


src/languages/loader.ts
```diff
@@ -0,0 +1,50 @@
...
+    languagesConfig[code] = configFunction(
+      { companyNames, ...translationOverrides[code] },
+      secondaryButtonMode,
+      cookieTable[code] || {},
+    );
+
+    return languagesConfig;
+  }, {} as Record<string, VanillaCookieConsent.Languages>);
+
+const languagesMap = {
```
literat, line 34: Please, place this definition before it is used. I asked myself, why is this not exported, and after that, I noticed, that it was used by the function before. For more info there is this ESlint rule: https://eslint.org/docs/latest/rules/no-use-before-define

literat:
I think you have a bad target branch here. It looks like this is based on the branch with the implementation of the language loader. 🤔

literat:
> > I think you have a bad target branch here. It looks like this is based on the branch with the implementation of the language loader. 🤔
> 
> Yes, because we needed the live preview, and Netlify is set to only create it when the target branch is main - so this was a quick workaround, I will retarget and rebase it once the #327 is merged.

Ok, then my suggestion in code is meant to be applied in the previous PR. 🤷

literat (approved):

src/languages/loader.ts
OndraM: This rule is enabled and it don't triggers an error. Sooo, is this really an issue? :thinking:

OndraM: :boom:

PR #338, opened 2024-08-22
Deps: Update prettier to 3.3.3 and add stylelint-prettier rules

OndraM:


.prettierrc.json
literat, line 1: You cannot just remove this file. There is also the `format` script in the `package.json` which reads this configuration.

Also the `stylelint-prettier` reads this configuration and uses it as a base.

literat: Oh, my bad, you are moving it to the `package.json` :-)

PR #355, opened 2024-10-14
Deps: Temporarily lock esbuild version

OndraM:
Esbuild 0.17+ break the `.serve()` API (must be rewritten to async context, which breaks other things). Until we rewrite this, keep the old version to allow development.

See https://github.com/evanw/esbuild/blob/main/CHANGELOG-2023.md#0170


OndraM:
Replaced by #356

PR #371, opened 2024-11-26
Use API of vanilla-cookieconsent v3 & refactoring

OndraM:
This PR will be failing until all CCM internals are refactored for the new version of vanilla-cookieconsent.


src/LmcCookieConsentManager.ts
```diff
@@ -77,95 +78,105 @@ const LmcCookieConsentManager: CookieConsentManager = (serviceName, args) => {
...
   } = options;
-  const cookieName = 'lmc_ccm';
-  const cookieConsent = window.initCookieConsent();
-
-  const onFirstAcceptHandler = (
-    userPreferences: VanillaCookieConsent.UserPreferences<CookieConsentCategoryValues>,
-    cookie: VanillaCookieConsent.Cookie<CookieConsentCategoryValues>,
-  ) => {
-    const cookieData = cookieConsent.get('data');
+  const cookieName = 'lmc_ccm'; // TODO: alma_ccm?
```
literat, line 81: **question (non-blocking):** Does this change during the finalizing of this PR or will this be done in the future? Maybe linking some issue will be helpful :-)

Same for the next TODO :-)

OndraM: This one is already implemented in follow up PR.

I added reference to the issue for the second TODO.

PR #382, opened 2024-12-10
V3 integration test [do not merge]

OndraM:


.github/workflows/markdown-link-check.yaml
```diff
@@ -10,8 +10,8 @@ jobs:
     name: "Markdown link check"
     runs-on: ubuntu-latest
```
literat, line 11: **suggestion (if-minor):** What about using a version number like in other workflows?

PR #388, opened 2024-12-17
Feat: Use deep merge for config overwrites (fixes #385)

### LINKED ISSUE #385: Deep merge `config` instead of override
With cookieconsent v3 the `config`  [became ](https://cookieconsent.orestbida.com/reference/configuration-reference.html) more structured. So there is basically nothing relevant to be overwritten on the first level, the interesting config is inside `cookie` key etc.

However, we now populate (overwrite) our default config with user-provided config with destructuring: `...config`. Meaning you must provide the complete structure of the multi-level object, otherwise you will loose all the other config on that level.

Possible solutions:
- [lodash merge](https://lodash.com/docs#merge)
- https://github.com/jhildenbiddle/mergician

OndraM:
This allows to overwrite multi-level deep config object of the vanilla-cookieconsent.

Used library [Mergican](https://github.com/jhildenbiddle/mergician) is light-weight and does not extend the dependency footprint, as it has no external dependencies.

### DIFF
--- a/src/LmcCookieConsentManager.ts
+++ b/src/LmcCookieConsentManager.ts
@@ -1,5 +1,6 @@
 import 'vanilla-cookieconsent';
 import { nanoid } from 'nanoid';
+import { mergician } from 'mergician';
 import submitConsent from './consentCollector';
 import {
   CategoriesChangeset,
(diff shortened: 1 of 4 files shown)


src/LmcCookieConsentManager.ts
```diff
@@ -125,50 +126,51 @@ const LmcCookieConsentManager: CookieConsentManager = (serviceName, args) => {
...
-    cookie: {
-      name: cookieName, // Predefined cookie name. Do not override.
-      expiresAfterDays: (acceptType) => {
-        return acceptType === 'necessary' ? 60 : 365; // 2 months or 1 year
+  const cookieConsentConfig: CookieConsentConfig = mergician({})(
+    {
+      autoShow: true, // Show the cookie consent banner as soon as possible
+      cookie: {
+        name: cookieName, // Predefined cookie name. Do not override.
+        expiresAfterDays: (acceptType: string) => {
```
literat, line 134: **suggestion (if-minor):** It would be more appropriate to have a more precise type than the string. Maybe a union type like `necessary | ...`. There should be a dictionary of accepted values, or not?

literat (approved):

src/LmcCookieConsentManager.ts
OndraM: Good point, I missed there is AcceptType type in vanilla cookieconsent. See fixup.

PR #390, opened 2024-12-19
BREAKING CHANGE: Rename callbacks

OndraM:
To match their new names in vanilla-cookieconsent library.

This causes BC, however, I find this better than having different names from the vanilla library.


literat (approved):
LGTM a make sense 👍

PR #396, opened 2025-01-02
BREAKING CHANGE: Rename settingsModalMoreInfo to preferencesModalMoreInfo

OndraM:
In CC 3.0 "settings modal" is now everywhere called "preferences modal". 

To make it consistent also in `translationOverrides`, we change `settingsModalMoreInfo` option to `preferencesModalMoreInfo` and also rename the relateD scss file.


literat (approved):
Naming is hard...

Issue #392, opened 2024-12-19
Rename remaining names to Alma Career


OndraM:
All functions, types and files should be renamed:

- [ ] `initLmcCookieConsentManager` -> `initAlmaCareerCookieConsentManager`
- [ ] `LmcCookieConsentManager` -> `CookieConsentManager`/`AlmaCareerCookieConsentManager`
- [ ] `LmcCookieConsentManager.ts` -> `CookieConsentManager.ts`
- [ ] `LmcCookieConsentManager.scss` -> `CookieConsentManager.scss`
- [ ] CSS variables (`--lmcccm` -> `--ccm`/`--acccm`/...?)

literat:
- [ ] Rename package scope from `@lmc-eu` to `@almacareer`

OndraM:
> * [ ]  Rename package scope from `@lmc-eu` to `@almacareer`

There is already separate issue for this, because its not related to the code itself: https://github.com/lmc-eu/cookie-consent-manager/issues/391

OndraM:
@literat @adamkudrna What new prefix (without the LMC part) for the css properties would you suggest?

Currently the naming scheme is `--lmcccm-font-family`, `--lmcccm-bg`, ...

So just remove the `lmc`, making it `--ccm-font-family`, `--ccm-bg`, ...?

adamkudrna:
> > So just remove the `lmc` part, making it `--ccm-font-family`, `--ccm-bg`, ...?
> 
> I am fine with this.

Why not. I was also thinking about `alma` (`--alma-font-family`) so the company name would be included. But I'm OK with `ccm`, no strong opinion.

OndraM:
Ok, so renaming the css properties is irrelevant (we [will use](https://github.com/alma-oss/cookie-consent-manager/pull/425) names from vanilla-consent), so closing this as done.

Issue #391, opened 2024-12-19
Migrate to almacareer npm org


OndraM:
The library should be migrated to almacareer NPM organization.

PR #397, opened 2025-01-02
BREAKING CHANGE: Rename LmcCookieConsentManager to AlmaCareerCookieConsentManager

OndraM:
Ref #392.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Load default CSS along with your styles in `<head>`:
 
 ```html
 <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Inter:wght@400;600;700&display=swap">
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@lmc-eu/cookie-consent-manager@2/LmcCookieConsentManager.min.css">
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@lmc-eu/cookie-consent-manager@2/AlmaCareerCookieConsentManager.min.css">
 ```
 
 Load the script and initialize the plugin right before ending `</body>` tag:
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -38,7 +38,7 @@ Load default CSS along with your styles in `<head>`:
 
 ```html
 <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Inter:wght@400;600;700&display=swap">
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@lmc-eu/cookie-consent-manager@2/LmcCookieConsentManager.min.css">
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@lmc-eu/cookie-consent-manager@2/AlmaCareerCookieConsentManager.min.css">
```
literat, line 41: **suggestion (if-minor):** 
```suggestion
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@almacareer/cookie-consent-manager@2/CookieConsentManager.min.css">
```

Two points:

- I think you forget to rename a package scope from `@lmc-eu` -> `@almacareer`
- Do we still need to use a prefix for the file? Isn't `CookieConsentManager` just enough? Because, I think, now the name of the file and the function names are way too long. And since this is a BC we can do better, IMHO.

OndraM: > I think you forget to rename a package scope from @lmc-eu -> @almacareer
It will be done in a separate issue, #391

OndraM: > we can do better

I'm not sure whether it is actually better. The idea behind the prefix was/is to clearly distinguish from the underlying cookieconsent library.


However, you are right the name is now a bit long (I wouldn't say too long, but definitely longer than before :smile: ).

I will prepare second PR without the AlmaCareer prefix, to see how it looks, and we may choose that one.

OndraM:
Replaced by #400

PR #404, opened 2025-01-09
Chore: Update version replace script to ESM

OndraM:
To fix

```
$ ts-node scripts/readme-replace-version.ts
TypeError: Unknown file extension ".ts" for readme-replace-version.ts
    at Object.getFileProtocolModuleFormat [as file:] (node:internal/modules/esm/get_format:183:9)
    at defaultGetFormat (node:internal/modules/esm/get_format:209:36)
    at defaultLoad (node:internal/modules/esm/load:116:22)
    at async ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:483:32)
    at async ModuleJob._link (node:internal/modules/esm/module_job:115:19) {
  code: 'ERR_UNKNOWN_FILE_EXTENSION'
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "version": "yarn changelog && yarn replace-version && git status && git add CHANGELOG.md README.md package.json ./dist && git status",
     "postversion": "echo 'Check and push: `git push --set-upstream origin main && git push --tags`'",
     "release": "yarn version --`./bin/ci/semver.sh`",
-    "replace-version": "ts-node scripts/readme-replace-version.ts",
+    "replace-version": "tsx scripts/readme-replace-version.ts",
     "types": "tsc -p ./tsconfig.json"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


literat (approved):
I have switched `ts-node` for `tsx` - https://tsx.is/, so you will have better DX with executing TypeScript scripts (do not care if CJS or ESM).

OndraM:
> I have switched `ts-node` for `tsx` - https://tsx.is/, so you will have better DX with executing TypeScript scripts (do not care if CJS or ESM).

Cool, thanks. It looks definitely better.

BTW it is always great to see how many ways to achieve such simple task as running a file there is in JavaScript ecosystem :-D

literat:
> BTW it is always great to see how many ways to achieve such simple task as running a file there is in JavaScript ecosystem :-D

The world is evolving 🤷

PR #408, opened 2025-01-20
Docs: Add catalog info for Backstage

OndraM:


catalog-info.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: cookie-consent-manager
+  description:  Alma Career Cookie Consent Manager
+  annotations:
+    github.com/project-slug: alma-oss/cookie-consent-manager
```
OndraM, line 7: (The repo will be moved to `alma-oss` before this PR is merged)

Mariyo: how big deal it could be to rename the repo according to [Software structure ADR](https://backstage.almacareer.tech/catalog/default/area/acc-engineering/adrs?record=1206f3e8-8bc2-80f5-b7b2-fe461bbbc320)? 
It could be `alma-oss/consents-cookieConsentManager` where `consents` is domain and `cookieConsentManager` component of type `library`

catalog-info.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: consents-cookieManager
```
Mariyo, line 4: `consents-cookieConsentManager` should be the name according to [Software structure ADR](https://backstage.almacareer.tech/catalog/default/area/acc-engineering/adrs?record=1206f3e8-8bc2-80f5-b7b2-fe461bbbc320)

Mariyo (changes requested):

catalog-info.yaml
OndraM: I updated it in the meantime to `consents-cookieManager`, based on suggestion from Marek. Is it OK?

catalog-info.yaml
OndraM: I would strongly oppose renaming the repository - I don't think it is meaningful to apply internal naming rules to open source projects. I thought this is part of the point for `alma-oss` organization - to have OSS projects (on which apply different rules) separated from internal projects. 

Open source projects have different priorities for their name: is should be clear, simple and "cool" looking - because they are intended for public presentation. So I would keep names like "cookie-consent-manager" and "spirit-design-system", and not align them with the internal software structure naming.

OndraM: The name combines dash and camel case notation, looks cryptic and it has unnatural English word order. It is not "product name" but rather "categorization name", which is good for organization of hundreds of internal projects, and I fully support that. However for a name of public open source "product" it does not make sense, because the name of the repository is also the brand of the project (which is something that is not the case for internal projects and internal repositories).

On top of that, we already for multiple years use the term "cookie-consent-manager" as name for the project and name of the NPM package. So it is generally expected for the repository to have the same name.

Furthermore, we have these "Alma OSS GitHub Organization Guidelines", which were agreed on by DEX team, and they allows this kind of naming: https://www.notion.so/almacareer/Alma-OSS-GitHub-Organization-Guidelines-1586f3e88bc28099a45ee95b9951c61b?pvs=4#1586f3e88bc28056837ffd9df218f220

literat (approved):
LGTM

catalog-info.yaml
Mariyo: Lets summarize what we know:
- in Alma Career we have convention to name repos according to [ADR](https://backstage.almacareer.tech/catalog/default/technology/github/adrs?record=d9aada9f-3f5b-4dbb-aac0-8fe242a4bd47)
- alma-oss will contain all OSS repos from Almamedia, not only Alma Career
- owner of alma-oss is @aripalo
- according to @martas-z-almamedia-career, there is a verbal agreement about “dns like” convention and relative freedom in naming
- there is request https://github.com/alma-oss/.github-private/pull/1#discussion_r1827574733 from @martas-z-almamedia-career  to explicitly define naming conventions

To me it means, that there is no repo naming convention to follow, but Alma OSS org owner can change it in future.

PR #429, opened 2025-02-26
Chore: Use alma's sharable preset for renovate

literat:
* switch from user defined scheduling and grouping to predefinet presets

I have finalized the migration of `@lmc-eu/renovate-config` package (which is now deprecated) to repository base renovate-config (https://github.com/alma-oss/renovate-config), which is prefered solution from the Renovate itself.

So, now the repository configuration can be smaller.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,45 +1,15 @@
 {
   "extends": [
-    "config:recommended",
-    ":pinOnlyDevDependencies",
-    "helpers:pinGitHubActionDigests"
+    "local>alma-oss/renovate-config",
+    "local>alma-oss/renovate-config:scheduleWeeklyBusinessHours",
+    "local>alma-oss/renovate-config:groupMajorProdDependencies(npm)",
+    "local>alma-oss/renovate-config:groupNonMajorProdDependencies(npm)",
+    "local>alma-oss/renovate-config:groupAllDevDependencies(npm)",
+    ":dependencyDashboard"
   ],
-  "ignorePresets": [
-    "group:monorepos",
-    "group:recommended",
-    ":prHourlyLimit2"
-  ],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -1,45 +1,14 @@
...
   ],
-  "ignorePresets": [
-    "group:monorepos",
-    "group:recommended",
-    ":prHourlyLimit2"
-  ],
-  "baseBranches": ["main"],
   "enabledManagers": ["npm", "github-actions"],
-  "branchPrefix": "dependencies/",
-  "commitMessagePrefix": "Deps: ",
```
OndraM, line 15: We need to keep this temporarily, it is enforced by LMC conventional commits rules which we still use.

renovate.json
```diff
@@ -1,45 +1,14 @@
...
-  "labels": ["dependencies"],
-  "schedule": ["after 9am and before 5pm on Monday"],
   "packageRules": [
-    {
-      "groupName": "major prod dependencies",
-      "groupSlug": "major-prod",
-      "matchManagers": ["npm"],
```
OndraM, line 22: Please make sure we don't want to have this configured for `github-actions` package manager differently. Currently these rules only applies to npm, but after using the shared rules, they apply to all package managers. Do we want this?

OndraM (changes requested):

renovate.json
literat: I have modify the current presets to use the package manager as a parameter, please see https://github.com/alma-oss/renovate-config/pull/10.

renovate.json
literat: Ok, I have reverted this change and refactor this with the new commitlint config.

PR #430, opened 2025-02-28
Deps: Upgrade `@alma-career/stylelint-config`

adamkudrna:
Upgrade our Stylelint config to the latest version.

However, recent Sass deprecation of mixed declarations collides with our Stylelint config order rule. Let's step back to Sass and fix it in our Stylelint config later.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "vanilla-cookieconsent": "3.1.0"
   },
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


adamkudrna:
Our Stylelint config is fixed in https://github.com/lmc-eu/code-quality-tools/pull/200.

literat (approved):

OndraM (approved):

